PR #11, opened 2015-06-24
Proposal for Better Doc Pipeline

dblotsky:


stevengill:
Nice! Definitely liking option 2 on first glance but will take a closer look soon.

kant2002:
Also willing to have better docs process.

Here my point of view as translator. I'm in favor of current process.

##### Negative points
Here just points, why I think that way, really willing to improve docs.

* I not see any value in the Jekill so far. I see Jekill just as website generator, which is seems to be not needed for us now. I don't familiar enough with it, so would like to see how it could be used to improve docs. 

* What kind of templates do you envision in the docs? Right now we have separate templates for languages, this is includes whole set of resources, HTML, CSS, JS. Could you highlight that in proposal.

* Amount of effort required to convert plain text to autolinking seems to be underestimated. What are alternative for that in the Jekill? If similar alternatives exists, how hash references inside docs would change on different language, right now we don't care about that. How would we reference sections inside docs in language-neutral ways?

* CrowdIn occasionally broke Markdown when links are explicitly specified. Thanks MS team for excellent docs, but your links to MSDN give me some pain :smile:
* Currently translation with CrowdIn is a pain, not because of translation process itself, but because of additional verifications which has to be performed
  - Autolinks generated in same places for different language (we have tool for that)
  - Verification that CrowdIn export (Jsut finish that tool)

 This will require to migrate such tooling to Jekill too (some maybe obsolete, but I suspect new tools would be required also)

##### Positive points
* JSDoc for plugin docs is very interesting, willing to adopt something like that. Right now we have completely drop translated docs for core plugins from website and they don't appear from NPM. Willing to bring it back to main website somehow.

* 100% for republishing edge docs on each commit.

* Want to automatically publish changed docs to CrowdIn on commit too. Having that give me ability to faster translate docs, so Russian translation would be in sync with edge, and no more two week delays in translation.

* Willing to have report about translation status of language sometime.

sosahvictor:
Hello  @dblotsky here my comments

If I don't mention any requirement here is because I agree we need something to fulfill it.

**Edit on GitHub**
This is not possible because a resource modified outside of CrowdIn will be overwritten. The proper collaboration on translated docs MUST be through CrowdIn to ensure proper documentation versioning. There is a special case in Russian documentation, Andrey has voluntereed to manually translate docs into Russian, but in the special case Andrey step out of this duty, the Russian translation will be overwritten by CrowdIn via the automated process.

**Doc versions**
Don't we already have this? You can select the version using the combo in the upper right corner in the docs page. I agree that not all versions are there, but still the edge one (5.1.1 as of now) is there.

**Localization**
Not sure I'm getting this right, but the current translation strategy is to translate to the following languages using CrowdIn: Chinese simplified, Chinese traditional, French, German, Italian, Japanese, Korean, Polish and Russian (this on is done externally per Andrey's request).

**API docs**
You can use CrowdIn APIs to get the docs https://crowdin.com/page/api

**Constraints**
We've been doing automated translation for several months now under a free scheme. CrowdIn and Microsoft Machine translator are kind enough to allow us to use their products for free for this project but we host docs in Apache infrastructure (can someone keep me honest here, please?).

**Sphinx**
I think we require MD files in order to show documentation on GitHub as well. Can someone answer this?
Also,  what's the difference between using option 2 with our current process?

**My position**
I agree with Andrey, I see no value with Jekyll vs what we already have. Moreover, the autolinking problems we've seen are indeed injected by CrowdIn's process and my little availability to work with the quality tools Andrey has built for us.

Publishing on every commit? I'm not sure, we publish new documentation per major release for tools, platforms and plugins. I don't think publishing to edge right after a commit is the right way to go even if such feature is not yet public in a release (I've seen some "don't push until JIRA #### is done" commit messages and some get lost in the PRs backlog).

I also agree we need a good search mechanism in documentation and analytics though. Would love to have these two for our users (sometimes I find myself struggling looking for a specific topic).

kant2002:
More suggestions which is important to me as user of different language and I could volunteer to implement.

**Allow deployment of separate docs releases**
Right now seems that when new version added, there need to update all versions of docs for them to be aware about new versions. I propose put data required to populate Version combobox in the separate file and use that file to populate combobox. For those who run without JS enabled, we could create static page with version selector. One doc at a time, faster docs releases.

**Simplify version selector**
Fix [CB-2619](https://issues.apache.org/jira/browse/CB-2619). This will show user that they see outdated docs. Trivial to implement after previous item.

**Allow faster switch to preferred language**
Show user that he could see version of same docs on different language. Since we use machine translation, sometimes useful to look at the original instead of translated text. Also novices could find solution in SO pointing to English docs and then switch to their native language.

dblotsky:
#### Jekyll

Jekyll has support for things like `{% include %}` and template inheritance, removing code duplication (of which we currently have a lot). For example: with Jekyll, adding analytics to all pages would be a *one-paragraph change*. It also has plugins for things like **full-text search** and automatic cross-referencing.

Jekyll would let us move non-translation code out of the Crowdin files as well, which will *reduce the injected errors*. All existing tools will work with Jekyll without any modification.

#### Edit on GitHub

Given that on Crowdin our source language is English, we can still enable the editing feature, but only show it in the English docs. Alternatively, we can at least have a "create GitHub issue" feature, that would allow people to easily submit requests for changes.

#### API Docs

By APIs, I don't mean Crowdin APIs. I mean automatically-generated docs for Cordova's plugin/tool APIs. That is, automatic code documentation. This is what JSDoc and joDoc and YUIDoc do.

#### Versions and Translation

Yes, translation and versions are already handled. I made the list of requirements just so we can compare different doc generation systems.

kant2002:
**Jekyll**
I will leave it alone :smile: since your examples is good documentation. With our current pace of writing docs, there no need in Jekill. If somebody would like to volunteer and bring a ton of text in our docs I would be the first to help in that.

The best way to docs generation, is to start with docs structure which we would like to implement. I want to speak in terms of what kind of value we bring to different group of users. If you feel that our structure should be improved, then let's talk about that, and they select tool for that.

**Crowdin**
The issues with CrowdIn would be always present in the same amount as now. It is due to way how CrowdIn extract translation strings from Markdown. All issues so far from my experience is from dealing with lists. So until we move our docs to different format, or will avoid lists we will have issues. See http://cordova.apache.org/docs/es/edge/guide_platforms_android_webview.md.html for examples.

**Edit on GitHub**
Easy to implement, but how we would manage user initialed changes. Before we write about signing process on Android some guy wrote short topic on that and send us [pull request](https://github.com/apache/cordova-docs/pull/212), nobody respond to it. I try to bring attention to that pull request, but without luck at that time. I doubt that we have flow of users willing to improve our docs. If we have rare changes, that's obvious that nobody will look at docs. Somebody with commit rights have to take responsibility for that.

**API Docs**
Yes, I do understand about which docs you are talking. I think it is highly valuable for developers, since this is simplifies their work in documenting plugins.

**Full-text**
I want to have full-text search on website, but that seems to be impossible task for us. How we would rank pages in search results? Why do you think such plugin would be better then Google/Bing Site search?
Implementing full-text on such small text base just hide our UX issues which we have. 

**UX issues*
For example try to find article on Windows 10. I translate it, but it was not easy to find it. Initially I look for it on the "Platform gudes" page in the Windows section, but it is buried inside "Windows Platform Guide". This is inherent problem to our docs, and if we provide better navigation tools for our user, our docs immediately become better. 

Another problem is seeing right version of docs. Most of the time users need latest one, but sometimes they need specific version (for example using 3.7.1 Android platform version). Issue was reported around two years ago. That means we don't have person which is responsible for docs.

Same could be said about the language selector. To switch to same version of docs, but different language you have scroll a lot.

I want to point that I willing to resolve any of the issues which I highlight above. Not just panic :smile:

nikhilkh:
Should this PR be closed now that we have implemented most of the recommendations here?

dblotsky:
Sounds good to me.

PR #16, opened 2015-09-21
Plugin docs gen proposal

daserge:
# Cordova docs-gen proposal

### Intro

Cordova plugins documentation currently live separated from source code so that sync issues can occur on commits.
Proposed solution uses [JSDoc](http://usejsdoc.org/) to generate a plugin API docs, while allows to wrap the generated docs into template (markdown with handlebars partials) with extra-details like Quirks or additional examples.

### Libraries/frameworks used

* [jsdoc-to-markdown](https://github.com/jsdoc2md/jsdoc-to-markdown/) with source code links support patch ([generated doc result](https://github.com/MSOpenTech/cordova-plugin-camera/blob/5001fe35a091a6846e3207fe70fb5502743cfcd3/README.md)).

**Summary instructions**:

* Put a template including license, header, install steps and extra docs (quirks) to `jsdoc2md/TEMPLATE.md`,
* Document `www` folder source code using JSDoc,
* [Mark all template extra-docs sections, which will be linked to as an <a> with `name`](https://github.com/MSOpenTech/cordova-plugin-camera/blame/5001fe35a091a6846e3207fe70fb5502743cfcd3/jsdoc2md/TEMPLATE.md#L96),
    * [Link JSDocs to the template' extra docs sections](https://github.com/MSOpenTech/cordova-plugin-camera/blob/5001fe35a091a6846e3207fe70fb5502743cfcd3/www/Camera.js#L126),
* Call `npm install` to install dev dependencies,
* Call [npm run gen-docs](https://github.com/MSOpenTech/cordova-plugin-camera/blob/5001fe35a091a6846e3207fe70fb5502743cfcd3/package.json#L45) to generate the documentation,
    * Use [npm run gen-docs-abs](https://github.com/MSOpenTech/cordova-plugin-camera/blob/5001fe35a091a6846e3207fe70fb5502743cfcd3/package.json#L46) along with setting `--base-url` argument to generate absolute source links instead of relative ones (this could be used with cordova-docs site).

[See cordova-plugin-camera docs as an example](https://github.com/MSOpenTech/cordova-plugin-camera/blob/5001fe35a091a6846e3207fe70fb5502743cfcd3/README.md).

### DIFF
--- a/proposals/PluginDocsGen.md
+++ b/proposals/PluginDocsGen.md
@@ -1,0 +1,39 @@
+# Cordova docs-gen proposal
+
+## Intro
+
+Cordova plugins documentation currently live separated from source code so that sync issues can occur on commits.
+Proposed solution uses [JSDoc](http://usejsdoc.org/) to generate a plugin API docs, while allows to wrap the generated docs into template (markdown with handlebars partials) with extra-details like Quirks or additional examples.
+
+## Libraries/frameworks used
+
+* [jsdoc-to-markdown](https://github.com/jsdoc2md/jsdoc-to-markdown/)
+
+Proposal is currently implemented for cordova-plugin-camera only: [generated doc result](https://github.com/apache/cordova-plugin-camera/blob/dcc81bfbe16244cb460d3c5da79fe6ec22795ce7/README.md).
+
+**Pros**:
+
+* A nice TOC helps in understanding what APIs are exposed,
+* A better structure for args with a table showing default values,
+* Hyperlinks for type definitions,
+* Literate programming makes sense - docs & code remain together,
+* A common plugin [template](https://github.com/daserge/dmd-plugin-cordova-plugin) that is shared among all cordova official plugins including basic details such as:
+  * How to install,
+  * Where to file bugs,
+  * Where to discuss design questions,
+  * Provide a pointer to slack plugins community,
+  * Buildbot badge.
+
+**Cons**:
+
+* Requires a generation script before every commit
+
+**Summary instructions**:
+
+* Put a template including license, header, install steps and extra docs (quirks) to `jsdoc2md/TEMPLATE.md`,
+* Document `www` folder source code using JSDoc,
+* [Mark all template extra-docs sections, which will be linked to as an `<a>` with `name`](https://github.com/MSOpenTech/cordova-plugin-camera/blame/5001fe35a091a6846e3207fe70fb5502743cfcd3/jsdoc2md/TEMPLATE.md#L96),
+    * [Link JSDocs to the template' extra docs sections](https://github.com/MSOpenTech/cordova-plugin-camera/blob/5001fe35a091a6846e3207fe70fb5502743cfcd3/www/Camera.js#L126),
+* Call `npm install` to install dev dependencies,
+* Call [npm run gen-docs](https://github.com/MSOpenTech/cordova-plugin-camera/blob/5001fe35a091a6846e3207fe70fb5502743cfcd3/package.json#L45) to generate the documentation,
+    * Docs generation script will also be called [automatically*](https://github.com/apache/cordova-plugin-camera/blob/dcc81bfbe16244cb460d3c5da79fe6ec22795ce7/package.json#L42) via [husky](https://www.npmjs.com/package/husky) precommit hook (*`npm install` is required).


nikhilkh:
Here's what I like about these docs compared to the previous version:
* a nice TOC helps in understanding what APIs are exposed
* A better structure for args with a table showing default values.
* Hyperlinks for type definitions
* Literate programming makes sense - docs & code remain together.
* We can probably have one template that is shared among all cordova official plugins and include basic details such as:
      - How to install
      - Where to file bugs
      - Where to discuss design questions
       - Provide a pointer to slack plugins community.
       - Buildbot badge.

One downsides:
* Requires a generation script before every commit

@purplecabbage @stevengill @axemclion  Thoughts on this experiment? Is it worthwhile to move all of our plugins docs to this format?

stevengill:
I think it is worth giving a shot. For this to work, we have to make sure the comments/jsdoc are being updated when apis are added/changed. Our plugins don't see a ton of change, so it hopefully won't be a big issue.

Might want to consider running generation script via npm prepublish or a git pre-commit hook to simplify the process further. 

What happens if someone edits the readme.md file manually?

The readme looks much nicer. 

Sharing template.md accross plugins is a great idea. 

I would suggest updating the installation portion to include installation from github + old id for older cli users. Something like https://github.com/apache/cordova-plugin-contacts#installation

csantanapr:
I think this is great !! I find my self trying to decipher our plugins APIs, and end up reading thru the source code to understand how to use it ;-p

I think the other major feature AFTER this get's implemented, I would like to see cordova plugins generate typescripts definitions in https://github.com/borisyankov/DefinitelyTyped

DefinitelyTyped is becoming very popular in the javascript community to allow editors and IDEs to do autocompletion, and also sniff the APIs definition not necessary the developer would be writing apps using typescript they could be using ES6 or ES5 and still benefit from the definitions.

cfjedimaster:
Can I ask, what is the meaning of "camera.getPicture extra docs"

daserge:
Thanks for reivew!
@stevengill,
> What happens if someone edits the readme.md file manually?

It will be overwritten by gen script, so a change will be needed to be done in the template or in the plugin code.

@cfjedimaster,
> Can I ask, what is the meaning of "camera.getPicture extra docs" 

The idea of this linking is to move large docs chunks to the template to reduce noise in the code. This will also make a plugin to take less place in terms of disk storage (not a big deal though taking into account frameworks and other assets size).  
In this particular case extra docs are additional examples and quirks.

cfjedimaster:
Kinda makes sense. I think the wording there could be changed. "Errata"
perhaps.

On Thu, Sep 24, 2015 at 9:51 AM, Sergey Shakhnazarov <
notifications@github.com> wrote:

> Thanks for reivew!
> @stevengill <https://github.com/stevengill>,
>
> What happens if someone edits the readme.md file manually?
>
> It will be overwritten by gen script, so a change will be needed to be
> done in the template or in the plugin code.
>
> @cfjedimaster <https://github.com/cfjedimaster>,
>
> Can I ask, what is the meaning of "camera.getPicture extra docs"
>
> The idea of this linking is to move large docs chunks to the template to
> reduce noise in the code. This will also make a plugin to take less place
> in terms of disk storage (not a big deal though taking into account
> frameworks and other assets size).
>
> In this particular case extra docs are additional examples and quirks.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/cordova/cordova-discuss/pull/16#issuecomment-142952540>
> .
>



-- 
===========================================================================
Raymond Camden, Developer Advocate for MobileFirst at IBM

Email : raymondcamden@gmail.com
Blog : www.raymondcamden.com
Twitter: raymondcamden

purplecabbage:
Personally I would prefer well written human generated docs. I think this proposal mainly moves the problem instead of fixing it  
I am always opposed to dictating one-way only methodologies, and the fact that we cannot simply modify the readme directly, need to regen on commits, and must do things according to jdoc ways are all non-starters for me. 
The jira/slack links will never change, so there is no benefit there ...
I would prefer we be stricter about accepting code changes that should affect docs, and require an update to docs if the code changes merit it.

daserge:
@purplecabbage, I agree with your points.
> we cannot simply modify the readme directly

We can still modify the template to highlight some details.
Additionally [some live reload technology](http://www.ryanthaut.com/guides/sublime-text-3-markdown-and-live-reload/) can be used while developing/changing the docs or API.

> need to regen on commits

[git pre-commit hook](https://github.com/typicode/husky) could be used to regenerate the docs automatically.

> must do things according to jdoc 

I think this is a price for having types hyperlinks, enum values, etc.

nikhilkh:
@csantanapr We already have typescript definitions for all Cordova plugins - https://github.com/borisyankov/DefinitelyTyped/tree/master/cordova/plugins. Though there is some duplication now. This sounds like an interesting project to use Typescript (t.ds file) instead of JSDoc for documentation and generate markdown docs from that. You get red of all the funky type referencing in JSDoc as a result. I don't see a tool that currently does this - sounds like a great intern project - I will suggest the Typescript team to do this.

daserge:
Published common DMD plugin for Cordova plugins:
[dmd-plugin-cordova-plugin](https://github.com/daserge/dmd-plugin-cordova-plugin)
[TEMPLATE.md now uses its partials with parameters](https://github.com/MSOpenTech/cordova-plugin-camera/blob/02cddcd03c89a0f9ea0cefe92955befddcdd34dc/jsdoc2md/TEMPLATE.md), [the plugin itself is added to dev-dependencies](https://github.com/MSOpenTech/cordova-plugin-camera/blob/02cddcd03c89a0f9ea0cefe92955befddcdd34dc/package.json#L52) and used as a [DMD CLI argument](https://github.com/MSOpenTech/cordova-plugin-camera/blob/02cddcd03c89a0f9ea0cefe92955befddcdd34dc/package.json#L46) ([npm package](https://www.npmjs.com/package/dmd-plugin-cordova-plugin)).

daserge:
Added [husky precommit hook](https://github.com/MSOpenTech/cordova-plugin-camera/blob/505f1a0441eb8f8e2ea07a56dbef46c6238957fc/package.json#L45) to automatically re-generate docs.
Although this still requires a developer to `npm install` first (and also on further docs template updates) so that the hook is symlinked to local directory - [updated the template to include this warning](https://github.com/MSOpenTech/cordova-plugin-camera/tree/505f1a0441eb8f8e2ea07a56dbef46c6238957fc#this-documentation-is-generated-by-a-tool).

daserge:
Hello guys,
I have sent a [PR](https://github.com/apache/cordova-plugin-camera/pull/129) with the first version of the docs reworked to be generated, which does not include source code links for now, please take a look.
I wanted to discuss where we should store [the common plugin template](https://github.com/daserge/dmd-plugin-cordova-plugin) - it currently lives at https://www.npmjs.com/package/dmd-plugin-cordova-plugin.

purplecabbage:
Is this matter resolved? 
I see that we are now[1] telling pull-request senders that they must modify jsdoc2md/ files ... I oppose this change to workflow.

[1] apache/cordova-plugin-camera/pull/131

daserge:
@purplecabbage, regarding common plugin template - it still lives in my npm account:
https://github.com/daserge/dmd-plugin-cordova-plugin

Regarding `jsdoc2md/TEMPLATE.md` modification and a change to the workflow:
If we decide to have a generated documentation, someone will have to generate it - either contributor or committer. I automated the generation process via pre-commit hook but it requires `npm install` to run first so the process is not fully automatic now.

We could get rid of `jsdoc2md/TEMPLATE.md` via moving the Quirks to the JS code but this will add too much noise in it and what is more it will bring platform-specifics into the platform-independent code.

Let's discuss what we can improve in this topic.

PR #27, opened 2015-11-03
Proposal to Remove the Cordova iOS Native Whitelist

shazron:


csantanapr:
Is CSP applicable for iOS8 and below and this is the level of protection that user will get in new proposal?

csantanapr:
If CSP doesn't work or is not applicable for iOS8 and below, I want to say that we want to remove whitelist only for ios9 and leave it for iOS8.

If iOS8 and below they get protected only using CSP then it might be goo enough

csantanapr:
Is this line "Developers still have to specify <access> tags to whitelist domains when they are whitelisted to in the CSP however."  only applicable to iOS9+, meaning that <access> tag don't do anything in iOS8 and below?

shazron:
CSP applies to iOS 8 and below. Reminder c-ios 4 only supports ios 8 and 9 now by consensus on the list.

Yes, with the proposed change the access tag only applies for iOS 9 and above. The access tag will only apply for ATS which is only available on iOS 9 and above.

csantanapr:
Thanks shaz, I would sugges to make it more clear the sentence that if <access> tag is present in config.xml it's totally for iOS8 and below, only applicable to iOS9 and higher.

And the access tag, the only it does is prep Info.plist with ATS entries and nothing more?

shazron:
Will do @csantanapr - I will update it. 
The <access> tag will be parsed by cordova-lib and add Info.plist entries *only*, nothing more -- correct.

shazron:
Text updated to reflect my discussion with @csantanapr

jessemonroy650:
@shazron
Sorry, I have not been able to get to this. The project I was working on in California fell apart on the way back to Texas, I moved forward on the whitelist system. I have comment on this. My opinion - "My sincere hope is that within the next 5-10 years we can get rid of, or replace, this whitelist system, including CSP". 

The largest part of the problem is they are both using regular expression. While this may seem like a good idea, it is in fact bad. I've taught RE (Regular Expression) to people for over 15 years, and it is a bad idea to use RE as a security filter. I will elaborate when I get some more time.

https://web.archive.org/web/20051026185025/http://www.accu-usa.org/2001-04.html

https://web.archive.org/web/20051026190312/http://www.accu-usa.org/2001-05.html

shazron:
@jessemonroy650 that's what this proposal is about, removing Cordova's own whitelist system, and leaving CSP in -- which is what partially what you want to happen. See the diagram. This is currently proposed for cordova-ios only.

shazron:
Consensus on the ML: remove it. Filed: https://issues.apache.org/jira/browse/CB-9972

shazron:
Merging.

jessemonroy650:
Thanks. I will get the word out.

jessemonroy650:
I'm working on a blog, but need to do a table or matrix. It's too confusing othewise. My raw notes on the blog are here: https://github.com/jessemonroy650/top-phonegap-mistakes/blob/master/whitelist-ios-notes.md   Does the table look correct?

jessemonroy650:
@shazron These boiler plate repositories need to be updated too:
- https://github.com/apache/cordova-app-hello-world
- https://github.com/phonegap/phonegap-start
I'm working on Phonegap, as this app is a constant source of problems with *Phonegap Build*. Beginners are constantly tripping over minor issues. I'll fix the references to plugins with this.
And add many more comments.

PR #28, opened 2015-11-06
Plugin Audit proposal

stevengill:


proposals/PluginsAudit2015/audit.md
```diff
@@ -0,0 +1,223 @@
...
+* Open Issues: 116
+* Open Pull Requests: 36
+* API needs a refactor
+* Cordova Webview based inappbrowser as a replacement would have big security issues if we opened up third party links in it. So we shouldn’t do it.
+* Carlos suggested new inappbrowser based on safariview. He can elaborate more.
+* encourage people to use it through apis instead of window.open shims
+* demo of oauth workflow + demo of closing inappbrowser via js (url change)
+* http://ngcordova.com/docs/plugins/oauth/
+* get rid of insertcss & executescript. Use fileapi instead
+* Recommendation: Keep. Needs a refactor. Remove insertcss & executescript. Encourage use through apis instead of window.open. Make demos doing oAuth. Maybe phonegap-plugin-oauth.
```
dpogue: Something based on SFSafariViewController (iOS) and Chrome Custom Tabs (Android) would be ideal. Some OAuth providers (Fitbit, for instance) have started adding restrictions that you're not allowed to do the OAuth flow in non-browser web views.

proposals/PluginsAudit2015/audit.md
```diff
@@ -0,0 +1,223 @@
...
+* Recommendation: Keep. Work on creating more consistency across platforms. Especially with cdvfile. Create phonegap-plugin-file with improved api.
+
+## File Transfer
+* 1 Month Downloads: 16597
+* Open Issues: 72
+* Open Pull Requests: 8
+* xhr2 is an alternative
+* available on all the platforms now. 
+* http://caniuse.com/#search=xhr2
+* Recommendation: Keep for now, but also start recommending xhr2 instead. Create tutorials using xhr2 instead of file-transfer.
```
dpogue: Worth noting that XHR2 requires CORS headers. I've had to use the File Transfer plugin in the past specifically to work around stuff that wasn't returning CORS headers.

proposals/PluginsAudit2015/audit.md
```diff
@@ -0,0 +1,223 @@
...
+* extra properties if cell data? 
+* Recommendation: Keep, investigate to see if extending functionality makes sense
+
+## Status-Bar
+* 1 Month Downloads: 33072
+* Open Issues: 27
+* Open Pull Requests: 3
+* no spec
+* pretty much required for iOS
+* Recommendation: Keep, add to hello world template
```
dpogue: Can this be handled without an explicit API by parsing meta tags for `theme-color`?

If the theme-color is solid, show the status bar in that colour. If the theme-color has alpha, tint the status bar that colour and overlay it on the webview.

proposals/PluginsAudit2015/audit.md
```diff
@@ -0,0 +1,223 @@
...
+## Media 
+* 1 Month Downloads: 6396
+* Open Issues: 68
+* Open Pull Requests: 19
+* Too many quirks between platforms
+* Maybe implement web audio api spec instead as a plugin
+* web audio supported in iOS 8.4+ and Android 5+
+* create web audio plugin for Android 4 to 4.4
+* alternative: https://www.npmjs.com/package/cordova-plugin-nativeaudio
+* Recommendation: Deprecate! Recommend using web audio api instead and look into creating a web audio api plugin to polyfill support for android 4 to 4.4.
```
devgeeks: The polyfill might also look after any native tweaks needed to make WebAudio better like background modes etc on iOS.

devgeeks:
Looking really good.

proposals/PluginsAudit2015/audit.md
```diff
@@ -0,0 +1,223 @@
...
+
+## File
+* 1 Month Downloads: 21252
+* Open Issues: 60
+* Open Pull Requests: 15
+* phonegap-plugin-file or phonegap-plugin-fs. Consumes cordova-plugin-file. Create downstream version with a better interface. Current one is based on standard which is ugly. (rip shelljs file ops)
+* Can we treat some urls consistency across all platforms? Possibly use Flex Air package for inspiration.
+* ex: app:, temp:, 
+* cdvfile wasn’t consistent enough across platforms
+* Recommendation: Keep. Work on creating more consistency across platforms. Especially with cdvfile. Create phonegap-plugin-file with improved api.
```
agrieve: Could use File wrapping in app-harness as one point of inspiration: https://github.com/apache/cordova-app-harness/blob/master/www/cdvah/js/ResourcesLoader.js#L149

proposals/PluginsAudit2015/audit.md
agrieve: It's also the case that when downloading to a file, XHR can't do so directly, so it has to go through blob -> file reader -> base64 string -> exec() bridge -> native -> binary -> disk. AKA, rather inefficient. :(

proposals/PluginsAudit2015/audit.md
stevengill: Thanks for the tip! 

downstream distro (PG file?) discussion wasn't supposed to be in here. Just removed that line

proposals/PluginsAudit2015/audit.md
```diff
@@ -0,0 +1,223 @@
+# Cordova Plugins
```
agrieve: Would also love to have https://github.com/MobileChromeApps/cordova-plugin-background-app be considered for folding into platforms. Many common patterns on Android don't work well without it (e.g. push notifications)

proposals/PluginsAudit2015/audit.md
```diff
@@ -0,0 +1,223 @@
...
+* Open Issues: 13
+* Open Pull Requests: 17
+* Using custom styled divs should be used instead of alerts
+* In HTML 5.1 there is a proposed dialog element which actually seems pretty good. http://www.w3.org/html/wg/drafts/html/master/single-page.html#the-dialog-element 
+* looks good on android
+* looks ugly on ios
+* currently not sync which is how default alerts work 
+* window.alert still available but says index.html
+* Recommendation: Deprecate! Encourage use of HTML/CSS/JS divs instead.
+
```
cfjedimaster: I'd argue against this. While it is common for UI frameworks to include a dialog element, being able to use the default device dialog is handy and should be kept. One line to show a dialog is *incredibly* useful compared to N lines of CSS and JS.

proposals/PluginsAudit2015/audit.md
```diff
@@ -0,0 +1,223 @@
...
+**NOTE**: Data for this proposal was collected on October 28th, 2015. 
+
+## Battery-Status
+* 1 Month Downloads: 5116
+* Open Issues: 18
+* Open Pull Requests: 3
+* Spec: http://www.w3.org/TR/battery-status/ 
+* android 5 has the ability to grab battery info in uiwebview built in
+* isPlugged is useful but not being used much
```
omefire: Just curious: how do we know which one of these API endpoints are used a lot and which ones aren't ?

proposals/PluginsAudit2015/audit.md
sgrebnov: Native background download/upload support may be interesting.

PR #29, opened 2015-11-24
Adding pinning and versioning proposal

stevengill:
Please review and leave feedback. Want to get this in Cordova 6.0.0

### DIFF
--- a/proposals/pinningAndVersioning.md
+++ b/proposals/pinningAndVersioning.md
@@ -1,0 +1,40 @@
+# Pinning and Versioning
+
+## Pinning Plugins
+
+We want to start pinning core plugins to `cordova-lib`. When a user using the `cordova-cli` runs the command `cordova plugin add cordova-plugin-device`, the version that is pinned in `cordova-lib` should be fetched from **npm** instead of fetching the `latest` tag.
+
+This assumes:
+* user has not specified a version during add step. (`cordova plugin add cordova-plugin-file@2.0.0`)
+* user does not  have a version specified in `config.xml`. (`<plugin name="cordova-plugin-device" spec="2.0.0" />`)
+
+The user specified version will always trump the pinned `cordova-lib` plugin version.
+
+I plan on pinning the plugins in the `package.json` of `cordova-lib`.
+
+    {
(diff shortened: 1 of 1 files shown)


axemclion:
I think that rather than pinning plugins in lib, a plugin should specify which semantic version of the CLI it depends on. Cordova-lib pinning plugin versions will only benefit core plugins, and we need a solution that also takes care of popular, non core plugins.

Hence, my proposal would be to continue to use the engine element in plugin.xml to enable the plugin to specify the minimum version of cordova CLI that it needs. Cordova-lib on the other hand could also walk the npm version lists of the plugin to figure our which version works best with the installed CLI.

stevengill:
This proposal doesn't plan on replacing the engine tag in plugin.xml. Right now the engine tag is used to specify which platform versions it supports. This will continue. Our tools already support if the engine tag specified a cordova version in the engine tag. So that is possible and a good idea. 

We will continue to recommend 3rd party plugin developers to use the engine tag. 

In terms of cordova-lib traversing npm versions to find the best plugin, I don't think this is a good idea. The engine information lives in plugin.xml. The only way for lib to potentially traverse through versions would be to fetch from npm, unpack it, read config.xml. If the engine tag fails, move on to the next version. Also, this is too much magic IMO. It makes more sense to say if you are using X version of cordova, you get Y version of a plugin and Z version of a platform. If you want a different versions, this is how you do it, but we don't guarantee it will work.

treggiari:
Just my opinion, but I think Steve’s proposal, and explanation for making the change, are perfect!

Leo

stevengill:
Sent a small PR for this. https://github.com/apache/cordova-lib/pull/345

macdonst:
@stevengill LGTM

I like that a user installing cordova will get a working app without having to immediately go an update a bunch of plugins. Users who want to over-ride which plugin gets installed have plenty of options.

riknoll:
Hey all, I realize that this was merged but I still think there is some discussion to be had on this topic. I've posted another discuss PR and would love it if people could review it. The PR is here:

https://github.com/cordova/cordova-discuss/pull/30

PR #31, opened 2015-12-07
New IA for Cordova website

rakatyal:


dblotsky:
Nice proposal! I like that it's splitting plugin and app development into separate buckets. Some comments:

1. I think Reference should be its own heading and not under App Development.
2. If we have tutorials, where should they live?
3. Should there be a section for developers for Cordova itself, which documents the components of Cordova and its architecture?

RobPosener:
Also needs a section on "Migrating from One Cordova Version to Another" as well as "Migrating from One Plugin Version to Another".  I see these as containing a complete historical set of instructions so that if developers skip a version they can still see what they need to do to migrate.

rakatyal:
@dblotsky: 1. Yeah I also had my doubts regarding that. But don't you think it would be better if we could also segregate references based on app or plugin developer? For example, an app developer might not be interested in plugin.xml at all and finding it inside a common reference may confuse him. Thoughts?
2. Right now we didn't create a separate heading for tutorials and added the relevant articles instead. But if we can create tutorials, I guess it would live under it's own top level heading.
3. Is there anything else that the current IA does not cover?

rakatyal:
@RobPosener: I believe we will cover that under 'Manage versions and platforms'

dblotsky:
1. My thoughts are still on a standalone reference. I can't imagine a user being confused that among the things in the Reference that they _were_ looking for, there might be some things they also _weren't_ looking for. However, I can definitely imagine them being confused when they go to docs looking for the CLI reference about how to add the _plugin_ they're developing, only to discover that they had to click on Create Apps to find it. That is, I think references should be scenario-agnostic, and that is why they should have their own heading. If I'm alone in this, then perhaps we should keep it as is. Should we vote maybe?
2. I can't see there being harm in having an independent Tutorials section to plan for the future. Same as above: vote?
3. Same as `3.` above: I think we should have a section for Cordova devs (and just curious users), documenting Cordova architecture. Almost everyone I've met has been confused about the separation between platforms and CLI.
4. What @RobPosener said: where will the upgrade guides be?
5. Right now the docs don't support multi-level ToC. Will this IA require them to support it?

rakatyal:
@dblotsky: 1. I understand your concern. We can go with the community's view point on this.
2. Yes we should add a stand alone section once we have the tutorials ready. 
3. I think we shouldn't add any unnecessary documentation. It might be confusing to users and hard to maintain.
4. There is already an article 'Manage versions and platforms' under Create apps section. It will be covered in that.
5. Yes. We would need a multi level ToC.

dblotsky:
1. Yeah, I'll bring it up on the mailing list and see what people think.
2. Cool.
3. As per some external discussion, there seems to definitely be a need for docs about Cordova architecture. I suggest that we rename "Overview" to "What Cordova Is", and that we add an item called "How Cordova Works" under the "Create Plugins" title or wherever appropriate
4. Cool.
5. Ideas of how to do multilevel ToC should probably then be in this or another proposal.

csantanapr:
I like the overall Information Development Architecture that is driven by user tasks.

As a app developers I want to do x for my app, here are the docs to achieve x
As a plugin developer I want to do x for the plugin I'm providing, here is the docs to achieve x

I think we should a few sections
1. for App Icons and Splashscreens under "create app"
2. Managing Plugins (add, remove, list,)
3. Customizing using Hooks
3. Sharing your app in source control, recommend when posting your source code do not share platforms and plugins, which can be restore.

I think we can have a "What Cordova is" that is focus more for user trying to understand what this apps.
Explaining more about using web dev skills, using js apis to access native capabilities.

Under References we can have a more in depth Architecture/Overview where it shows how everything is connected.  We can devide this into "Tooling Architecture" and "Runtime Architecture"

I agree "References" should be at the top level in TOC, since it applies to both app and plugin builders.

kant2002:
I would like to suggest structure based on some suggestions from Carlos and my earlier thought on that subject. I try to give reasoning why I create each of topic and for whom I think it would be interested
Here the audiences which I think of when create such structure:

1. Architects and experienced developers who willing investigate Cordova as platform for movign forward their mobile efforts. For that kind of people it is interested to see, how Cordova working, what are limitations, and in general more about machinery.
2. Novices Developers who just want to start their first Cordova apps, Web developers who want to wrap their existing SPA to Cordova, etc. These people just want get start fast, and then tinkering later with what happens.
3. Developers who what to learn how to do specific common development tasks in Cordova way.
4. Developers who already familiar with Cordova and want extend it functionality with some plugin.
5. Superhumans who know how to integrate Cordova in their existing applications.
.
Topic
- Overview
- Architecture (Audience: Architects who investigate technology)
	- Runtime Architecture
	- Tooling Architecture
- Create apps (Audience: Developers who want start writing apps with Cordova)
	- Create your first app
	- Develop for platforms
	- Managing re
	- Debug and test
	- Publish to stores
	- Automate Tasks
	- Manage versions and platforms
	- Manage security
		- Whitelisting
		- Security best practices
		- Manage privacy
- Accessing device functionality (Audience: Developers who would like to add new functionality to their existing apps)
	- Store data
	- QR & barcode
	- NFC
	- Network communications (even if this is jQuery, we have to show that jQuery could be used, File transfer is a bit of tricky, Whitelisting should be covered here as well)
	- GPS functionality
	- Handling device rotation
	- Other stuff.... (lazy to expand)
- Create plugins (Audience: Developers who want start create plugin for accessing functionality of the phone, not covered by existing plugins)
	- Create a plugin 
	- Register with NPM
	- Develop for platforms
	- Use Plugman
	- Debug and test
- Advanced topics
	- Embed Cordova in native apps
	- Passing data to embedded Cordova app (This area not really covered, and that's why I never try to embed Cordova)
	- Passing data from Cordova app to the host
	- How to manage source code (This is highly opinionated, but if provide at least one way, then embedding would be easier)
	- Customize build using hooks
- Sample Apps
- Reference (Reference information for all Cordova functionality. I prefer have docs on plugins on the main docs website as well, since it is hard to navigate in GitHub docs, they are more API notes then docs now)
	- Config.xml
	- Events
	- Plugins
	- CLI
	- Hooks
	- Plugin.xml (I remove from Create plugins section, since current problem of Cordova that information are scattered across whole docs)

rakatyal:
@csantanapr : 1. We decided to include the icons and splash screen content inside 'Create your first app' article since this is a very common scenario. We would be changing the content from the current article.
2. Covered in 'Create your first app' as well.
3. Do we need anything beside a reference? We would be introducing it in the create plugin and app articles and linking it to the detailed reference article. 
4. This sounds like a good tutorial to have?

I agree with having a generic 'What Cordova is' and detailed tooling and runtime reference articles.
Thanks for the input!

rakatyal:
@kant2002: Sorry for the late reply. I like the idea to create an 'Architecture' and an 'Advanced Topics' section in the IA. 
Does it makes sense to move the 'Accessing device functionality' section inside the 'Create apps' heading? 
For the plugin docs, I think we should pull the content from the github on the website. I can ask the community to take the final decision on this.

rakatyal:
Updated the PR based on suggestions.

nikhilkh:
Shall we update this PR with status of what's done and what remains - and merge it in.

PR #35, opened 2016-03-29
Create Making Cordova Developers Happy

purplecabbage:
This is Rob Posener's discuss thread pr, which was not showing up.


brodybits:
Thanks @purplecabbage for taking the care to keep this discussion going. I am concerned that a number of people are affected by the problems described here. I will add some more comments when I get a chance.

One concern is that this discussion is now based on a comment on a commit within a pull request. This could easily be lost, turning into "comment on an old diff" by possible mistake.

purplecabbage:
true, but we need a forum to discuss this.  I would not object to doing so on the dev mailing list, but it should not be in the 1 year old 65+ response thread.

brodybits:
I would vote for a new issue which includes or quotes the text of the comment with the formatting cleaned up. Then members of the community can go through and pick out the root causes.

Now I will rant a little bit.

My take is that Cordova became an excessively complex project/platform/framework with no obvious place where people can get started.

What I think we really need is to have a more centralized place where people can easily get started, learn how to do one thing at a time, start with some simple plugins, try some more complex plugins, and start building bigger and bigger apps.

For example, the sqlite plugin that I maintain is one of the more complex plugins. So far I have been able to keep most of the documentation in its README. I have a simple explanation of what it is about, how to get started, how to install, how to test the installation, how to run it with a simple program, etc. etc.

One of the worst parts is that PhoneGap (which we know was a predecessor to Cordova but was kept alive as a distribution) was supposed to go away over time. Now it has started to become its own industry!

It would be great if we could have a more obvious and centralized place where people learn what Cordova basically is (a poly-fill, a framework to run HTML5/Javascript apps on mobile and other devices, a bridge to provide native functionality, and tooling to enable developers to develop, test, and package mobile apps written in HTML5/Javascript technology), how to get started with simple apps, how to use basic plugins, how to use third-party plugins, how to get started with writing a custom plugin, etc. etc. In addition, there should be a guide for troubleshooting, what to do before asking for support (try one thing at a time to determine what exactly is going wrong), what information to include when asking for help, etc. (I have added README sections and other guides for the sqlite plugin.)

We know that Cordova is largely supported by volunteer efforts in many areas, and it has been largely thankless work on the part of some really wonderful maintainers and other members.

Just my $0.02 worth.

cfjedimaster:
"My take is that Cordova became an excessively complex project/platform/framework with no obvious place where people can get started"

Well, I think it is also important to remember (and telling users) that app development, in general, is complex as well. While we can strive to make things as painless as possible, at some point, the end user has to accept that building a good, performant, mobile app is not something they are going to do in 5 minutes. At some point, Cordova has to say, "Ok, now go out on your own" as it can't possibly cover the full depth of - well - everything. That's where books (disclaimer - I have one), online courses, presentations, etc, can help fill the void. It's a bit like saying Microsoft's SQL Server product docs should be the end all of everything. I don't expect that and I don't think our users should expect it either.

"One of the worst parts is that PhoneGap (which we know was a predecessor to Cordova but was kept alive as a distribution) was supposed to go away over time. Now it has started to become its own industry!"

Really? I was never under the impression that it was meant to go away. Adobe has done their own thing with it - adding their own services, etc, and their own additions to the CLI. This is similar to what Ionic has done, and heck, even the MS TACO stuff. I don't see this as a problem per se. When I teach Cordova, I focus *just* on it to help keep folks focused, but obviously bring up PG, Ionic, etc.

"It would be great if we could have a more obvious and centralized place where people learn what Cordova basically is...."

I think the Next Steps guide (http://cordova.apache.org/docs/en/latest/guide/next/index.html) provides direction there (and - again - I'm biased as I wrote the first draft ;) - but I do think parts of it could be expanded into longer docs, especially the debugging. That's covered in Next Steps, but you may not know to look there.

"what information to include when asking for help"
Could that be covered by a pinned note in the Cordova Google Group? For JIRA, can that information be included in the New Issue form?

kerrishotts:
> "what information to include when asking for help"
> Could that be covered by a pinned note in the Cordova Google Group? For JIRA, can that information > be included in the New Issue form?

@cfjedimaster: 

We have a pinned note on the Google Group that mentions things we need and the like. I haven't seen anything similar for the new Adobe Forums, but it would probably be a good idea to add something there (IMO).

shazron:
I cleaned the formatting up, but Github is giving me an error when I edited the original comment by Rob Posener, so I put it up in a gist for now: https://gist.github.com/shazron/22ad196952c5b749aa074a35fee46f4d

shazron:
Added a new PR with the doc: https://github.com/cordova/cordova-discuss/pull/36

shazron:
Any objections if I close this and make everyone comment on #36? It already has a reference to this PR anyway

RobPosener:
No objections from me.
I like your editing... it makes it a lot easier to read.
Rob

kerrishotts:
@shazron: No objections; It's much easier to read with your formatting. :-)

purplecabbage:
Closing!

shazron:
Not sure who is subscribed to #36, but I added responses to that.

PR #36, opened 2016-03-31
Added Making Cordova Developers Happy doc by Rob Posener

shazron:
Rendered: https://github.com/shazron/cordova-discuss/blob/making-cordova-developers-happy/proposals/MakingCordovaDevelopersHappy.md

### DIFF
--- a/proposals/MakingCordovaDevelopersHappy.md
+++ b/proposals/MakingCordovaDevelopersHappy.md
@@ -159,7 +159,9 @@ I know that the list below is long and doesn't have many details.  I could write
 
     **Recommendation 27:** Develop more automated tests on complex apps, rather than just simple tests.
 
-20. Cordova development cycle times: The overall Cordova development operation comes across as Cordova contributors writing good code, even fantastic code, but not supporting it or delivering what Cordova app developers need.  Yep, new features and bug fixes are being churned out in a very fast manner (the “train” as I have seen it called), but as the features have increased, the number of regression bugs have increased.  Recommendation 28: Slow down the pace of releases.  
+20. Cordova development cycle times: The overall Cordova development operation comes across as Cordova contributors writing good code, even fantastic code, but not supporting it or delivering what Cordova app developers need.  Yep, new features and bug fixes are being churned out in a very fast manner (the “train” as I have seen it called), but as the features have increased, the number of regression bugs have increased.  
+
+    **Recommendation 28:** Slow down the pace of releases.  
 
     **Recommendation 29:** Spend more time performing integration testing before publishing a new release.  
 
--- a/proposals/MakingCordovaDevelopersHappy.md
+++ b/proposals/MakingCordovaDevelopersHappy.md
@@ -1,0 +1,186 @@
+# Cordova Contributor
+Rob Posener
+
+## Background
+
+This email thread is very important but seems to have “died”.  I want to contribute by providing some first-hand app developer experience and recommendations to fix this situation.  I have tried to limit my comments to MY specific experience with Cordova and leave it to you to extrapolate (or not) to “there may be others seeing and feeling this”.  As an overall comment, its NOT just the quality of the documentation that lets Cordova down, its a combination of many things.
+
+From the tone of some of the comments in this email thread, I'm not sure that anyone is going to listen, let alone act on what I have to contribute here.
+
+From the tone of a number of responses that I have received when I have tried to contribute (yes, I have signed an ICLA), I’m not sure that anyone is going to listen, let alone act on what I have to contribute here.
+
+I have been a software developer since 1965.  I have also held most IT roles.  I have used many tools and languages.  I have been developing software using HTML, CSS and JavaScript since 2000.  I have been using Cordova since 2013.
+
+Since 2013 I have used Android Studio, Visual Studio, the CLI and UltraEdit as tools in conjunction with Cordova.
+
+My vision for what Cordova does/should be (rightly or wrongly – but generally set by the “hype” surrounding Cordova) is:
+
+1. To allow me to code once and release many times to different platforms.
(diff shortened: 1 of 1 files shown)


shazron:
Some nomenclature for the discussion. 

For "Recommendation X", I will just abbreviate it as "RX". e.g. `Recommendation 1` will be `R1`. 

For each numbered point in the section ["Some Examples of Where Cordova has Gone Wrong for Me"](https://github.com/shazron/cordova-discuss/blob/making-cordova-developers-happy/proposals/MakingCordovaDevelopersHappy.md#some-examples-of-where-cordova-has-gone-wrong-for-me), I will just abbreviate as "PX". e.g. Point `2.` will be `P2`.

Since there are 35 Recommendations, I will be chunking them up and addressing them 5 at a time.

shazron:
**R1, R2, R3** - These are all important, we could tackle these recommendations after the other recommendations are addressed.

**R4** - this is addressed by the following comments.

**R5** - I think some of this work will be incorporated by the browser platform since it will not implement most of the plugin APIs, and this work should be added to the core plugin code when necessary.

shazron:
**R6** - OK. (dependent on R5)

**R7** - We would need to do a documentation review, and it should be paired with a unit test for each API function published.

**R8** - I will write issues for this once I understand the issues involved specifically for the splashscreen plugin. 

**R9** - We wanted to avoid duplicating the docs, especially since the W3C File API has gone through a lot, and we didn't want the docs to go to waste if we 'upgraded' the API, but it appears that we are stuck with what we started with and we should document this.

**R10** - OK. Good idea.

shazron:
**R11* - [CB-10844](https://issues.apache.org/jira/browse/CB-10844) I still can't repro. Might be a miscommunication about how it works, we just punt off the search to Google Site Search. We're improving this in [CB-11104](https://issues.apache.org/jira/browse/CB-11104)

**R12* - [CB-10872](https://issues.apache.org/jira/browse/CB-10872) was already filed, so this is being addressed.

**R13* - We could document these contracts, but currently any contributor/plugin author can circumvent these through hooks. We've decided not to impose any "hard" security limits on the folders of a project, but we could revisit.

**R14* - Please clarify with an example?

**R15* - Perhaps elaborate with a specific issue for the TACO team to build on.

shazron:
**R16** - See R7, R8. I believe we address this in the Quirks sections in the docs.

**R17** - I believe we are working on this or want to work on this, I can't find the discussion or issue on this yet though, will edit this once I find it.

**R18** - See R17 above.

**R19** - Suggestion? RSS comes to mind since that is already built in to JIRA. We don't have any other mechanism currently besides a dev@ email blast. To notify devs, they would have to opt in -- or are you saying a mechanism through the cordova-cli?

**R20** - Triage is being done by everyone. The contributors employed by their respective companies have their own project managers and priorities, but we do try to coordinate work whenever we can.

shazron:
**R21** - Not sure how we can improve this, or if this is available in JIRA itself.

**R22** - Filed [CB-11106](https://issues.apache.org/jira/browse/CB-11106)

**R23** - Filed [CB-11107](https://issues.apache.org/jira/browse/CB-11107)

**R24** - See R20.

**R25** - See R20.

shazron:
**R26** - Is your case about Android specifically? In general it's slower because we do a re-build each time.

**R27** - I'm not sure if we can cover every case here, but our mobile-spec does pull in each core plugin and runs all the tests while all the plugin code is loaded (js and native).

**R28** - Disagree about pace, agree about better tests and coverage so we don't have to worry about regressions.

**R29** - Agree, see R27.

**R30** - Agree, also see R27.

shazron:
**R31** - Don Coleman wrote [this](https://github.com/don/cordova-plugin-ble-central) which supports iOS and Android, and is a Cordova committer. We could ask him if he wants to contribute his code.

**P22** - Edit PR and Link to email thread “Edit to Edit on Github for Docs”.

**R32** - Filed [CB-11108](https://issues.apache.org/jira/browse/CB-11108).

**R33** - OK.

**P24** - I agree that communication can be terse sometimes, and JIRA is our "todo" list for all our work so we can keep it organized. See R34 for a possible solution.

**R34** - Apache JIRA has a web form that can be embedded so users can file issues without needing a login, we could add that to the website. Not sure about the "email to file issue" feature whether that is available in Apache JIRA.

**R35** - That would be ideal, but it has a lot of moving parts, see R20.

nikhilkh:
**R9**: We just merged a commit that adds samples for common use cases for the File plugin: https://github.com/apache/cordova-plugin-file/commit/5caa768a6b7eebd67ceaab10035e90740eee2c5d
**R15** Please elaborate on this - the Microsoft team has been improving support on this. I will try to follow up with you later to see specific issues that you are facing here. One of the biggest things that will make your app automagically compatible is setting `windows-target-version` preference to 10.0 as it uses Edge for its rendering which is significantly more compatible to Safari & Chrome than IE is.
**R16** For splashscreen on Windows, this has been addressed recently with this commit: acb45f3a386bbb80d1d52cbe9d26f3de2617bfe0
**R18** I believe we support this for plugins already - but it's a cool idea to do it for platforms. However, one version of the app ideally already encodes what version of platform/plugin it works with.
**R20** I triage every week and go through every JIRA that's filed. I agree we could do with communicating top issues/regressions.
**R24** Perhaps an FAQ would be useful.
**R26** We are actively working on this. These PRs should help: Faster Android build: https://github.com/apache/cordova-android/pull/299, Incremental prepare: https://github.com/apache/cordova-lib/pull/429. We have more ideas which we'll share soon.
**R28** Perhaps we can do release candidates.
**R31** There is also this one: https://github.com/randdusing/cordova-plugin-bluetoothle which works pretty well from our testing

shazron:
It's been about 6 months without a response from the original reporter, so I will resolve this.

PR #41, opened 2016-04-22
Adding Android N discuss

riknoll:
This discuss gives the results a brief investigation I did on the upcoming Android N changes that affect Cordova.

### DIFF
--- a/proposals/android-n-support.md
+++ b/proposals/android-n-support.md
@@ -1,0 +1,46 @@
+# Android N and Cordova
+
+The purpose of this discuss is to highlight the new features of Android N
+that may affect Cordova. I did a brief investigation of the upcoming changes
+and did some minor testing (building a trivial app, running mobilespec, etc).
+I didn't find too many issues, but we should track them as they come up.
+
+You can see a JIRA query for all Android N related issues
+[here][android-n-query]. If you file an issue, be sure to add the "AndroidN"
+label so that it shows up.
+
+
+## Multi-Window Support
+
+This is the big new feature of Android N and one that presents some minor
+problems for existing Cordova apps. When apps that are compiled with the pre-N
+SDK are put into multi-window mode, the user gets a message saying "App may not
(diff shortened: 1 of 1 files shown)


infil00p:
After reading the blog post, this doesn't surprise me.  I just wish that Android would have docs on this instead of just blog posts on stuff that you have to somehow be lucky enough to see on twitter or G+ the right day.  This is very similar to orientation changes on Android and we shouldn't have a problem supporting this.

However, given that people do pixel-perfect designs instead of reactive mobile designs (despite me telling them to do otherwise for years), I can see people wanting to turn this off.  I'm fine with just adding this in.  This is relatively minor compared to Marshmallow permissions.

riknoll:
Yep, I don't think it should have much impact. We just need to make sure that people are aware that they need to update their apps to either turn off multi-window or add support. Or at least that they know to try it out.

proposals/android-n-support.md
```diff
@@ -0,0 +1,46 @@
...
+just destroys/restarts the current Activity whenever the app changes size.
+Unfortunately, in a Cordova app the webview gets destroyed along with the
+Activity and the app will appear to restart from scratch. In practice, this
+mostly looks like the app randomly restarts sometimes when it's resized. The
+app also sometimes crashes (may be related to the first point
+[here][other-behavior-changes]).
+
+The fix is a really simple one-line change to AndroidManifest.xml and it works
+even with pre-N SDK versions. We just need to make sure that people are aware of
+the issue so that they can either update to newer versions of cordova-android (once
```
nikhilkh: Should we go ahead and make this change or wait for more Android N builds to do it?

nikhilkh:
LGTM. Let's merge this in.

proposals/android-n-support.md
riknoll: I can open a PR now, but it isn't urgent and it would probably be best to wait for [preview 4](http://developer.android.com/preview/overview.html) or so.

PR #43, opened 2016-04-29
Telemetry Proposal

omefire:
The goal of this proposal is to suggest an implementation of telemetry in cordova.

### DIFF
--- a/proposals/Telemetry.md
+++ b/proposals/Telemetry.md
@@ -46,6 +46,11 @@ Users shall be shown a timed prompt (with a link to a privacy notice) the first 
 If the prompt times out with the user having not provided any answer,
 it will be assumed they do not agree to telemetry being collected.
 
+### No Telemetry Flag
+Users might opt out of telemetry on a command basis by using the ```--no-telemetry``` flag.
+That way, they have the ability to enable telemetry and yet still disable it for specific commands as they see fit.
+e.g: ```cordova run ios --no-telemetry```
+
 ### Continuous Integration environments
 In CI environments, there will be a way to disable the telemetry prompt on first usage by setting an environment variable.
 In case the environment variable is not set, the command invoked will wait for a few seconds, then proceed. This should ensure existing tools that make use of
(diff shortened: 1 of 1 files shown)


purplecabbage:
This looks great!
I like the fact that we are doing this in cordova-cli and not cordova-lib so that different toolchains that use cordova-lib as a module will not skew our data.  I wonder if we should also include a cli arg for any tooling that use cordova-cli directly via shell. --donottrack or something

omefire:
@purplecabbage tools that depend on cordova-cli directly will set an environment variable: 
https://github.com/omefire/cordova-discuss/blob/Telemetry/proposals/Telemetry.md#continuous-integration-environments

purplecabbage:
First prompt aside, I think there needs to be a switch to turn it off per command.
Isn't it possible that I could have a globally installed instance of cordova-cli that I use frequently and I have allowed to collect telemetry data, but I also have another tool cordova-paramedic which may collect it's own telemetry data, and uses the globally installed cordova-cli. (actually, I know it's possible because I do ;) )
cordova-paramedic interacts with cordova-cli via `exec('cordova ...` and I would expect there are other tools that do too.
cordova-paramedic (and other tools) should be given a way of calling cordova cli commands without logging them. 
While `cordova logging on|off` seems like it could help, since we don't know what it was, we cannot restore it when we are done.

csantanapr:
Our IBM mfpdev CLI execs Cordova CLI having a global argument for every CLI command will be helpful

omefire:
@purplecabbage @csantanapr I can see a use case for a ```--no-telemetry``` flag to be used to switch telemetry on/off based on single commands ( e.g: ```cordova run ios --no-telemetry``` ), I'm wondering if it's a scenario many people will be faced with. That said, I'm not opposed to such a flag.

omefire:
@purplecabbage and @csantanapr : the `--no-telemetry` flag has been added to the prototype code:
- https://github.com/apache/cordova-cli/pull/247/files#diff-8e2e5cc1e3a35d5820c648b58677adfbR106
- https://github.com/apache/cordova-cli/pull/247/files#diff-d9d9214113dea1c364abd4ba3c70be11R100

Please, help review that code.

purplecabbage:
lgtm!

PR #45, opened 2016-05-10
Adding Android Studio Project Structure proposal

infil00p:


stevengill:
Thoughts @nikhilkh @csantanapr?

nikhilkh:
@infil00p Can you comment on what's the compatibility impact of this change - to plugins and to cordova app developers? I'm not familiar with the differences between current Android studio project and our cordova-android structure.

csantanapr:
Big +1
Common use case for our enterprise customers they like to create a Cordova app and then switch to open with Android Studio and launch the app from there and any debugging they have todo there.

infil00p:
@nikhilkh This would change the location of the www directory, the AndroidManifest.xml and the application code and move it into a module called app.  This would then allow us at a later point to treat plugins like library projects more easily, and would allow us to be more compatible with Android Studio.  This is basically to fulfill the same use case that @csantanapr is running into as well.

infil00p:
Update: I managed to migrate the build.gradle from the old structure to the new structure, but it SHOULD work with Environment Variables and other configuration files that Cordova developers are used to setting for their environments/CI.  I'm sure that I'm forgetting stuff, because I have no idea about everything we claim we support, only what we actually support.

At any rate, more testing of the proof-of-concept branch would be awesome.

purplecabbage:
Is this done? Can we close this?

janpio:
JIRA issue is resolved, cordova-android 7.0.0 added the discussed changes.
Can probably be closed or merged (?).

PR #48, opened 2016-05-31
Update ThirdPartyPackageManagers.md

juliascript:
Propose new "spec" attribute for the framework tag in plugin.xml

### DIFF
--- a/proposals/ThirdPartyPackageManagers.md
+++ b/proposals/ThirdPartyPackageManagers.md
@@ -19,7 +19,7 @@ Proposed support is through the existing `<engine>` and `<framework>` tags.
     <engine name="nuget" version=">=3.2" platform="windows" />
 </engines>
 
-<framework src="path/to/mypod.podspec" type="podspec" spec="~> 1.1.2"/>
+<framework src="name_of_cocoapod" type="podspec" spec="~> 1.1.2"/>
 <framework src="path/to/mynu.nuspec" type="nuspec" />
 ```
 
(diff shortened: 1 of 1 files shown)


proposals/ThirdPartyPackageManagers.md
```diff
@@ -32,3 +32,14 @@ The `name` attribute would specify the command-line package manager to be used,
 ## Framework Tag
 
 The `src` attribute can be a relative path or a URL. The `type` attribute specifies the type of framework that corresponds to the `engine` supported. In this case type `podspec` means that the lib should be installed using the `cocoapods` package manager, and `nuspec` for the `nuget` package manager.
+
+## Spec Argument
```
macdonst: I'm not necessarily against this but the spec attribute would not be used for Android. When specifying a gradle package in android the framework tag looks like:

```
<framework src="com.android.support:support-v13:23+" />
```

where the last bit `23+` indicates what version of the Support v13 library to use. In this case version 23 or up.

Would we include a spec attribute for Android so that the tag becomes:

```
<framework src="com.android.support:support-v13" spec="23+" />
```

Mind you we'll still have to support the old way for backwards compatibility.

stevengill: We don't need to standardize this as it is platform specific. For iOS, it makes sense to have it separate as that is how it appears in the pod file. For gradle, it appears as `com.android.support:support-v13:23+`.

macdonst: Like I said, I'm not against how it is documented below. I just wanted to mention the difference and get our decision documented somewhere.

PR #49, opened 2016-06-09
added new template proposal

stevengill:


stevengill:
@csantanapr thoughts?

stevengill:
So I was chatting with @carynbear about this proposal.

Some things we should chat about adding to this proposal:

* Is now a good time to deprecate copy-from?
* is `--link-to` needed?
* can we drop support for `config.xml` living inside www directory? See: https://cordova.apache.org/docs/en/latest/config_ref/

stevengill:
@nikhilkh @purplecabbage @vladimir-kotikov please take a look when you have time.

ryanjsalva:
I'm sorry, I don't understand why we wouldn't want to copy over package.json. It seems like you would want package.json the majority -- not the minority -- of the time. Can you help me understand why we're special casing it here? It seems like we have the potential to create more confusion by adding a set of hidden exceptions.

csantanapr:
Steve LGTM it covers what we discussed on files to ignore that are not for the template and are part of delivery in npm

Also the name and version fields in package.json like config.xml it's a new idea I like it +1

purplecabbage:
Hi Ryan, 
The reasoning here is that in a template repo, there are some things that belong to the template-repo and there are some things that belong to the application that is created from the template.

ie. 
template-repo/.git/ <= this would be expected to be part of .gitignore, but if installed from a local fork, we still have an issue.
template-repo/README.md
template-repo/package.json
template-repo/template-src/package.json

Ultimately here, I think the issue is that our template logic is doing a blind copy of everything.  I think the target template itself should define what is part of the template, and what is not.



@purplecabbage
risingj.com

stevengill:
@ryanjsalva we definitely want `package.json`. The issue is we don't want to copy over the `package.json` being used to publish a template to npm since it will have incorrect values for name, version, keywords, author, dependencies, etc. We will copy `package.json` (and everything else) from a template defined subdirectory.

stevengill:
Here is a PR by @carynbear enhancing template support. 

Things to note:
- `--template` now skips copying the templates `package.json` as well as other files mentioned in proposal. 
- if a template defines a subdirectory, it will copy everything from that sub directory (including `package.json`)
- if a `package.json` is included in the subdirectory, it will be copied and updated. Specifically, `package.name` will be a lowercased version of the third argument `cordova create` takes (which is the human readable name) and `package.version` will be set to `1.0.0`
- the version in `config.xml` will also be set to `1.0.0`
- Removed symlinking code. This means `--link-to` won't be supported in `cordova-create` anymore. This didn't work well on Windows since it required administrative privileges. The code for it was very ugly and IMO not very maintainable. The usecase we lose is symlinking a www directory to your cordova app. I believe @carynbear has a branch where she has kept the symlinking code intact. What are your thoughts on this? 
- `--fetch` is accepted when doing cordova create. This will use `cordova-fetch` to grab module/repo and fetch it to your `~/.cordova/node_modules` directory to be copied over.

Please review it @csantanapr @nikhilkh @purplecabbage 

Anything to add @carynbear?

stevengill:
Oops, forgot to link to the PR. https://github.com/apache/cordova-lib/pull/456

stevengill:
Going to merge this in today

purplecabbage:
Wait please.

stevengill:
Merging in now!

vladimir-kotikov:
Hey @stevengill, @carynbear - there is some failures (see [here](https://ci.apache.org/builders/cordova-android-osx/builds/2878/steps/creating-mobilespec-app/logs/stdio)), caused by the change in apache/cordova-lib#456. Looks like `--copy-from` option in 'createmobilespec' is being ignored. Could you please confirm if this is just a bug or `--copy-from` is not supported anymore?

stevengill:
Thanks @vladimir-kotikov. I'll take a look

stevengill:
@vladimir-kotikov I pushed a fix to createmobilespec.js. it was using --link-to which we removed. I updated it to use --template

PR #51, opened 2016-07-21
CB-11607 added cordova-create breakout proposal

stevengill:


brodybits:
Would a project created by cordova-create honor the hook scripts of the Cordova plugins?

stevengill:
@brodybits create would copy over any hooks from the template but not run them. For adding platforms, plugins, building, hooks runner, running and preparing, you would still need cordova CLI or cordova-lib.

cordova-create updates fields like name and version in `config.xml` and `package.json`. It creates the cordova directory structure. It also handles fetching a template (via cordova-fetch) and using it to populate the cordova project.

stevengill:
Actually, I want to change the proposal to keep cordova-create in its own repo. cordova-lib repo is getting pretty hectic with all of the modules living there.

vladimir-kotikov:
+1 for separate repo! Having multiple modules in one repo sounds like an anti-pattern

TimBarham:
Should the config stuff get its own module first (or perhaps live in `cordova-common`)?

stevengill:
which config stuff? global config path? config.xml? local config.json?

PR #52, opened 2016-09-21
Deprecation of node.js 0.x and upgrade to a modern node.js in Apache Cordova

shazron:
Rendered: https://github.com/shazron/cordova-discuss/blob/node-0.x-deprecate/proposals/NodeJsDeprecation.md


kerrishotts:
Totally OK with this. I hope devs are using newer versions anyway, but if they absolutely need an old version of node, there's always nvm (or equivalent).

stevengill:
In favor of depreciation! Why skip node 4.x support?

shazron:
@stevengill Mainly two reasons: 1) Oct 1st 2016, node 6.x will be the default recommended version for users to download, with node 7 as "current" 2) enables us to use ~100% of the modern features of JavaScript (ES6 and ES7)

My assumption is, Cordova is used in a desktop environment so this will have minimal impact for people to upgrade if need be (perhaps in server environments they might be constrained, but then there are the node versions managers out there that will fill this role).

ryanjsalva:
The telemetry I've seen indicates that a large number of developers continue to use Node 4.x -- especially in enterprise environments where Cordova is popular today (e.g. customers of IBM, Telerik, Microsoft, Kony). I'm all in favor of deprecating support for 0.x, but **strongly** advocate that we more gradually deprecate Node 4.x.

mbraude:
What is the technical argument for this? Customers run different versions of Node for all sorts of reasons. Shouldn't we try to accommodate them rather than purposely break them? Is there a specific bit of functionality that Cordova wants to consume that only exists in Node 6?

shazron:
@mbraude  thus this proposal is up for discussion, thus it was a PR. *Nobody* was talking about this until I put up a PR, at least this is a start. I wanted to hear all arguments for/against.

shazron:
@mbraude regarding Node versions, at a certain point, people will have to upgrade. A deprecation period is reasonable, or staying at node 4 until a certain point like @ryanjsalva suggests, but staying at a certain version forever is not.

shazron:
@ryanjsalva what would be a good deprecation period? how long do enterprise customers need to migrate usually? 6 months or 1 year?

sosahvictor:
Hi @shazron. My 2 cents.

Deprecating v0.x seems the way to go as this is going out of the market quickly but I concur with @ryanjsalva. For the time being, v4 is still supported by the Node community hence I don't see why we should drop it. Moreover, I can see a lot of questions coming this way asking why Cordova isn't working on Node v4.

My take is that Cordova continues to support Node v4 until the end of the maintenance period which is April 2018[1], perhaps making a deprecation statement before that in favor of v6 (LTS at that moment)

One thing to keep in mind is that in the last year lots of security vulnerabilities have been found in Node and enterprises look for the latest version with such fixes in the stream they are currently using (v4 as of now and who knows their plans to move up to v6).

[https://github.com/nodejs/LTS#lts_schedule](url)

ryanjsalva:
@sosahvictor's approach seems reasonable. The only thing I'd add is that we may want to re-evaluate when the deprecation notice goes out. We probably want some kind of threshold. For example, I'd think twice about deprecating Node 0.x if 99% of our users were still using it. Which brings up a good question: do we get Node version number in our current Cordova telemetry?

panarasi:
Adding to the discussion, here are the numbers from the telemetry we added to the CLI. While only 27% of people use telemetry, it could be a representative of what people do on the CLI.


| Version | Count | 
----------|---------
v4.5.0 |  1,337
v4.4.7 | 546
v6.5.0 | 455
v6.4.0 | 264
v6.3.1 | 201
v4.2.6 | 179
v4.4.5 | 177
v6.2.2 | 127
v0.12.7| 116

Data from https://datastudio.google.com/#/org//reporting/0B-Ja5cNOX_XLTElHdWd4V2NUem8/page/xEE. 

This could help us decide one way or another.

ryanjsalva:
Thanks, @panarasi. That data suggests roughly 2/3 of Cordova developers are using Node 4.x. Given this evidence, I think we'd be good stewards of the project pursuing the strategy @sosahvictor recommends. Are there other factors I'm not considering?

jcesarmobile:
+1 to drop 0.x support
-1 to remove 4.x support

TimBarham:
Yeah, I'd definitely be in favor of dropping 0.x support.

From a dev perspective I'd love to be able to start using some of the features that 6.x provides, but I agree the numbers show us we have to maintain support for 4.x for now. 

So I'll add my +1 to dropping 0.x but continuing to support 4.x.

And @shazron - thanks for initiating this long overdue discussion!

shazron:
I'm drafting a timeline graphic to attach to the post regarding what we have discussed thus far, with the deprecation and drop of support milestones in it, for discussion. Hopefully we can come up with a "formula" like the Node LTS Working Group so we don't need to discuss this again when the time comes to move on from node v6.

One rule that I think we need to come up with, re: keeping node v4 support -- is that we only recommend what is the current LTS of node going forward for our new users. In this case, on Oct 1st 2016, it will be node v6.

shazron:
FYI, I added the redirect http://telemetry.cordova.io to the telemetry page that @panarasi linked to

shazron:
See the timeline graphic below. 

In summary:
1.  node 0.x will be deprecated from Oct 1st 2016 - Dec 31st 2016, and node 0.x support is dropped on Jan 1st 2017 (3 month deprecation period).
2. node 4.x will be deprecated from April 2017 - Mar 31st 2017, and node 4.x support is dropped on April 1st 2018 (12 month deprecation period).

cordova-node-deprecation-timeline

stevengill:
Thanks for putting this together @shazron! Looks great! 

It is sad that node 4.x has such low es6 support. See (http://node.green/). It would be nice to start using es6 in our development. Maybe we do that anyways and add babel to our build process?

kerrishotts:
Timeline looks good to me (thanks @shazron); given the high number of users on 4.x, it only makes sense to keep 4.x support (though I am a little sad regarding ES6 support).

shazron:
I believe the latest timeline I made solves the concerns of @sosahvictor, @ryanjsalva, and others, and I will update the proposal to reflect.

sosahvictor:
+1. Thanks @shazron

shazron:
Updated the PR with the timeline changes. I'll post this to the dev@ list next week, and create a blog post (since Oct 1st is this Saturday).

ryanjsalva:
@shazron, in practice, I suspect this timeline will address my concerns, but I want to clarify that I was actually advocating for a slightly different trigger to deprecate support. Rather than set a calendar date, I've found it helpful to establish triggers grounded in user behavior. So, for example, we might deprecate support for Node 4.x once fewer than 20% of our users are using older versions. Fortunately, now that we have telemetry in the CLI, this is possible 👍 .

Again, I suspect that by April 2018, there will be sufficiently few Node 4.x developers to justify deprecation, but it's worth us confirming in, say, February 2018 that our predictions still hold. Does that sound sane?

shazron:
@ryanjsalva I am trying to balance the needs of the project (contributions) and the enterprise, I believe waiting 18 months from now to drop node v4 is reasonable, even if the telemetry numbers don't support it (which I think it will by then). But, let me explain...

In April 2018, when node 4 is already end-of-life'd, it is already 33 months since its birth (almost 3 years). At that time, node 8 is already the active LTS (with node 6 entering the start of maintenance for 12 months), node 9 is branched, and node 10 will be coming down the pipe in 6 months (and will be active LTS in Oct 2018). (We only really care about even numbered versions in node).

By then, our active node requirement (v6) is already almost two major versions behind.  Users will have either downloaded a newer node, and have moved on, and we don't really want to require devs to downgrade to a really old node (the interim node v4 that basically is the merger of iojs and nodejs) / keep using ES5 to keep developing on Cordova, especially since they can't even use modern ES6/ES7 features that will be the norm by then. This is detrimental and not attractive to contribution.

That being said, yes, let's revisit it when the deadline of April 2018 nears based on telemetry -- I'm confident that the numbers will support removal of node v4 by then.

We're giving a huge berth so people will have time to upgrade from v4, but for future versions, should that 20% number be a hard and fast rule to start to remove support for a node version? 

@panarasi Does datastudio support aggregating the node versions by major version number?

shazron:
Ember.js Node LTS Support: http://emberjs.com/blog/2016/09/07/ember-node-lts-support.html 
"To make it easier for us to deliver on our support promise we have adopted a policy that the HEAD of our master branches will be compatible with all versions of Node.js currently receiving support." thus when its out of maintenance, they drop support for that node version.

From Laurie Voss, COO of npm: https://twitter.com/seldo/status/781258069678391296
"80% of users are on node 4.x or higher".

ryanjsalva:
@shazron, I hope I didn't mislead you. I intended to convey that I agree with your proposal and timeline. Thanks, though, for doing all the extra homework to help us both (a) gain greater confidence in the the approach and (b) communicate it to the community. I think you're doing the right thing.

shazron:
Thanks everyone! Filed https://issues.apache.org/jira/browse/CB-11946 and blog post PR will be in the issue, as well as posted on dev@

PR #53, opened 2016-09-24
added move save/restore to package.json proposal

stevengill:


dpogue:
Can we take this opportunity to separate the restoring from the preparing? There have been a number of unintuitive edge case bugs around restoring and preparing, and it seems nobody knows exactly how it's supposed to behave and every potential improvement is also potentially breaking someone's expectations of how it works. The current code is rather inefficient (calling the prepare step several times for each platform).

A new command like `cordova install` could restore platforms/plugins from package.json, while `cordova prepare` would just prepare the installed platforms/plugins without trying to restore.

TimBarham:
Hey @stevengill - could you maybe add a section describing the motivation for this change - what are the benefits?

mbraude:
Hey @stevengill, I'm concerned about this proposal because it will break existing projects when they migrate to new Cordova versions. That's also going to be difficult for tooling vendors like us, who support multiple Cordova versions through a single project template. What is your backwards compatibility story? It seems reasonable to me that you would change Cordova to write to package.json by default (so that new projects will have the new behavior, perhaps unless they opt out) but still respect config.xml if the settings are there. That way you can move people forward without breaking existing projects.

axemclion:
I think it would be a good idea to refer to why we are doing it, as @TimBarham says. Maybe we should also refer to the face-to-face minutes where we started discussing this proposal.

As I understand, the primary reason for doing this is to eliminate all our custom logic for saving and restoring, and simply depend on npm to do the right thing. This is also closer to ReactNative's plugin save/restore model .

One question I has was about the `<configfile>` we added in `config.xml`, from plugin.xml. How would that work? In config.xml, we had a way to add stuff to platform xml files. Will that still be a part of config.xml ? If the goal here is to eventually eliminate `config.xml`, where would that move ?

stevengill:
I will add a motivation section. But as @axemclion mentioned, it is to allow us to dump a bunch of custom logic + using a more modern/expected dependency system. 

To be clear, I don't plan to remove `config.xml` from the project by cordova@7. Some things like `<config-file>` and `<edit-config>` don't make sense removing from `config.xml`. 

I will do some research on all of the available tags/preferences we currently use in `config.xml` and we can see which ones would make sense to move over. 

@mbraude my plan is for there to be a automapping of information from `config.xml` to `package.json`. So when older projects update to cordova@7, we would automatically move the relevant information out of `config.xml` and into `package.json`. I want upgrading to be seamless, but we will have to  throughly test. Of course we may not hit every use-case, but that is why this is going to be a major update to cordova.

stevengill:
@dpogue I think that sounds like a worthwhile idea. I think it would be better to make it into its own proposal to get feedback about that change from all interested parties. I definitely think it would be related to this proposal.

ryanjsalva:
Hi, everyone! Long-term, I love this plan and I'm totally in favor of moving to package.json. My only concern is the timeline. Frankly, if the proposal is to remove support for config.xml in Cordova 7.0, I suspect that gives downstream distributors only a few months to prepare. If downstream distributors can't accommodate the new architecture on a compressed timeline, thousands of developers are unlikely to benefit from the new platforms that come pinned with Cordova 7.0. 

In the interest of brainstorming alternatives...

* Could we consider dual support for package.json and config.xml for a longer time period?
* Is it possible for new platforms to continue working with older versions of the CLI?

I don't want to unnecessarily slow us down, but I also want to make sure Cordova continues to work as a platform for the downstream distributors who make-up a sizable percent of it's use.

stevengill:
I'd love to get a bit more info from @ryanjsalva and from other IDE maintainers + downstream distributors as well.

Are IDE's using `config.xml` to restore platforms/plugins their own way? As in, are they not using cordova (shelling out/api) to platform/plugin add/rm and restore. If that is the case, I can see why they would need a bit of time to update to using package.json. But if they are using cordova under the hood, things should function as normal. I'd love to hear how your IDEs will break.

Unless we see changes in PlatformAPI or big changes in common, I don't see why platforms would stop working with cordova@6. Of course, a platform may do a major update themselves that require some new feature in cordova@7, but nothing that i am proposing here would be a breaking change for platforms. The issue would be that the pinned versions of platforms in cordova@6 would grow stale, but users could still get newer versions by being explicit `cordova platform add android@6`

I'm willing to consider dual support for a longer period, but I would hesitate to call it cordova@7 as it won't be a major change at that point. The major breaking change part is the dropping of saving plugins/platforms in config.xml

I haven't set a timeline for cordova@7 yet. I will create a new issue to track the various things we would like to see land in cordova@7 and then we can figure out the timeline. My early estimate would be early 2017 as a ship date. But it depends on what we decide to include.

TimBarham:
Visual Studio certainly directly manipulates platform and plugin information in `config.xml` in lots of ways (in particular, in ways that there is currently no Cordova API for). Going forwards maybe we can review all the ways we do that and see if it makes sense to introduce `config.xml` / `package.json` agnostic APIs on the Cordova side to do it for us.

purplecabbage:
Is it possible to cut this proposal in half? I have concerns about dependency management differences between npm modules and our own plugin dependencies.
Plugins do NOT depend on each other in their own package.json, but they do in plugin.xml.

npm modules each maintain their own nested dependencies, and are able to do so without conflict because of the way they are loaded/resolved.  Our runtime plugin dependency system is a completely different beast and doesn't roll in easily.

If we could limit this proposal to just include platforms and tooling I would feel better about it. Perhaps as step 1, then we could evaluate if we wanted to continue onto the next level.

purplecabbage:
Another item that may be of concern is the general performance of npm in windows.
I have heard indications from Microsoft that a large number of first time cordova builds fail in windows because of some npm issue, be it firewall, network, a bad module or some other issue. We need to be sure that we are not making this worse as more that half of our users are in windows.

stevengill:
@purplecabbage we are not removing our managing of plugins dependencies. Plugins would not move the <dependency> tag from plugin.xml over to `package.json`. 

Top level plugins + their deps would be added as dependencies to the project's `package.json`. But the logic of adding plugins, adding plugin deps and removing plugins + their deps would stay the same. 

At its bare bones, this proposal is essentially just moving over our current dep saving from `config.xml` to `package.json`. 

You can literally test this out now by including a `package.json` in your cordova app and running the following. 

```
cordova plugin add cordova-plugin-device --save --fetch
cordova plugin add cordova-plugin-file-transfer --save --fetch
```

with the above code, `cordova-plugin-file` will be added as a dep in our projects `package.json`. It would not be listed as a dep in `cordova-plugin-file-transfer's package.json`.

stevengill:
`node` and `npm` are **required** to be installed on your system to use cordova. If users have poor `node` & `npm` setups, we will have to point them to documentation to fix their setups. The [fetch proposal](https://github.com/cordova/cordova-discuss/blob/master/proposals/fetching.md) which is implemented behind the `--fetch` flag for now, uses the users system installed `npm`. This is will be replacing our existing method of `npm cache add` to some hidden `.cordova` directory living in the home directory. 

local plugins + platforms will still work as an alternative to `npm`.

dpogue:
Just realizing one thing we would lose is the ability to pass variables to plugins:

```xml
    <plugin name="cordova-plugin-customurlscheme" spec="~4.2.0">
        <variable name="URL_SCHEME" value="ca.dpogue.my-app" />
    </plugin>
```

stevengill:
@dpogue Yes! You are right. I'm going to think about how to handle it. Any suggestions?

stevengill:
I'm thinking something like this:

```
"cordova": {
    "platforms": ["android", "ios"],
    "plugins": ["cordova-plugin-device", "cordova-plugin-compat", "cordova-plugin-customurlscheme"],
    "cordova-plugin-customurlscheme": {"name": "URL_SCHEME", "value": "ca.dpogue.my-app"}
}
```

Thoughts?

purplecabbage:
I think we still need to keep some data in config.xml, so to me it is acceptable to store plugin variables there in the same way.

Assuming you just ran : 
`cordova plugin add cordova-plugin-customurlscheme --save --variable URL_SCHEME=ca.dpogue.my-app`

It would result in this addition to package.json : 
```
"dependencies": {
        "cordova-plugin-customurlscheme": "~4.2.0"
}
```

and in config.xml :
```
<plugin name="cordova-plugin-customurlscheme">
  <variable name="URL_SCHEME" value="ca.dpogue.my-app" />
</plugin>
```

The only difference in config.xml is that the spec is not stored, and only plugins with variables will appear here.

stevengill:
@purplecabbage and I just had a in-depth discussion about this. Thought we don't agree on everything, the rough consensus between us seems to be:

package.json => for saving and restoring
config.xml => for preferences & available at runtime (if you are crazy enough to parse xml at runtime lol).

plugin preferences example: ```<splash src="res/screen/ios/Default~iphone.png" width="320" height="480"/>```

Saved variables are designed to be only used for restore. See http://cordova.apache.org/docs/en/6.x/config_ref/index.html#variable. So with that in mind, it makes sense for it to live in `package.json` in the `cordova` key.

I'm thinking something like this, but I'm still considering different combinations of objects/arrays
```
"cordova": {
    "platforms": ["android", "ios"],
    "plugins": ["cordova-plugin-device", "cordova-plugin-compat", "cordova-plugin-facebook4"],
    "cordova-plugin-facebook4": {
            "APP_ID": "123456",
            "APP_NAME": "myApp"
    }
}
```


Of course, until **cordova@8**, variable information would be stored in both `config.xml` & `package.json`.

Will be sending out a revised proposal today

proposals/save-restore.md
```diff
@@ -0,0 +1,97 @@
...
+* `cordova plugin/platform add/rm --save` will add/rm the plugin/platform to the new `cordova` key. If your plugin has variables, they will also be added/removed to the `cordova` key in this step.
+* restoring:  Look in both `config.xml` and `package.json` for plugin and platform dependencies. Use `cordova` key from `package.json` to restore. Restoring currently happens on a prepare. Restoring may become its own command in **cordova@7**. Read the proposal for that [here](https://github.com/cordova/cordova-discuss/pull/5://github.com/cordova/cordova-discuss/pull/54). 
+* autosave: remove functionality. Have docs telling users they can set autosave in npm config if they want it. `npm config set save=true`. `cordova-fetch` will use system *npm*, so this *should* work, but we will have to test it.
```
dpogue: Need to make sure that this can be detected by cordova so that the `"cordova"` key is also updated.

stevengill: good point

purplecabbage:
do we plan on continuing to support these flags?
```
--searchpath=<directory>
--noregistry
--force
```

I would also like us to do a full trace of what will happen for each of the following user inputs.  What will be the change in package.json and config.xml

```
cordova plugin add cordova-plugin-device --save
cordova plugin add https://github.com/apache/cordova-plugin-device --save
cordova plugin add ../cordova-plugin-device/ --save

cordova plugin add cordova-plugin-device --save --browserify 
cordova plugin add https://github.com/apache/cordova-plugin-device --save --browserify 
cordova plugin add ../cordova-plugin-device/ --save --browserify 

cordova plugin add https://github.com/apache/cordova-plugin-device --save --link 
cordova plugin add ../cordova-plugin-device/ --save --link
```

mlynch:
At first glance I like the approach of using a custom key in package.json and I think the format makes sense.

Thinking more deeply, many modern JS tools like bundlers and such are inferring modules they understand straight from `dependencies` or `devDependencies` and then looking for custom keys in their package.json's, like `module` or `jsnext:main`.

Basically: is it possible or does it make sense to just allow people to `npm install --save cordova-plugin-camera` and then cordova treats the plugins it finds in the standard npm modules as its own? I think that would be ideal, but then you have to infer...

purplecabbage:
Unfortunately @mlynch we would need to know ahead of time, every plugin id that exists.  
At the package.json level (via `npm i --save`) we can't even tell a platform from a plugin.  Technically it might be possible to run a postInstall script in the plugin itself that did the extra work, but this would require a change to all plugins.

stevengill:
Thanks for joining the conversation @mlynch! 

I have thought about this as a potential solution.

**Assuming we add a key to every plugin + platform `pacakge.json` to identify if it is a plugin or platform**: I seriously considered this. This would require us to read the `package.json` of every module in your apps `node_modules` directory. With *npm@3* having flat dependencies, this could mean we have to do hundreds of `package.json` lookups to see if something is a plugin or platform. An expensive operation but doable since this usecase is only for restoring (not frequently used). The bigger issue with this plan is requiring every platform and plugin to have a special identifier in its `package.json`. All existing plugins would break. I'm more nervous about this. The `package.json` lookup would have to happen during prepare (restore)  or its own command (`cordova install`). So a step after `npm install` would still be required.

The keys would have been `cordova:plugin` & `cordova:platform`.  I think these keys are good to start adding now since they help with discoverability on npm anyways.

**Assuming we use module name to identify** : We can't rely on module names as a way to identify what is a plugin or platform. Plugins and platforms can be named anything. 

**Assuming we use post install scripts in plugins + platforms**: As @purplecabbage said, this would require every plugin & platform adding a specific post-install script that would do the remaining portions of `cordova platform/plugin add` automatically on `npm install plugin/platform`. Not viable and too flakey IMO. 

I would love to have `npm install cordova-plugin-camera` work out of the box. I just haven't seen a way that doesn't require every plugin + platform in our ecosystem make a change.

mlynch:
Makes sense, given that cordova needs more data, it's probably best to be explicit. 

Random thought: One thing that could make traversing the module tree easier down the road is if plugins are `import`'ed or `require`'ed on the JS side and there is a module loader involved, as you'd at least know what plugins to go and look at, but that also feels like it wouldn't be a ubiquitous use case. And I'm not sure how it would solve the other issues. Something for the future perhaps

stevengill:
That is a great suggestion. Didn't think about it but it is something we could possibly use down the line. In **cordova@7** the plan is to start using **browserify** for plugin JS loading. I'll save this topic for the **cordova@7** proposal thread once I get around to writing it.

stevengill:
@purplecabbage not a full trace, but adding comments:
```
cordova plugin add cordova-plugin-device --save --fetch //works, added in both package.json and config.xml
cordova plugin add https://github.com/apache/cordova-plugin-device --save --fetch //works, added in both
cordova plugin add ../cordova-plugin-device/ --save --fetch //only added in config.xml. Need to switch fetch logic to support local paths

cordova plugin add cordova-plugin-device --save --browserify --fetch //works
cordova plugin add https://github.com/apache/cordova-plugin-device --save --browserify --fetch //works
cordova plugin add ../cordova-plugin-device/ --save --browserify  --fetch //only added to config.xml

cordova plugin add https://github.com/apache/cordova-plugin-device --save --link //--link only supports local paths
cordova plugin add ../cordova-plugin-device/ --save --link //no fetch support for local paths yet, we shall see. Currently it saves the local path in config.xml and symlinks the plugin
```

`--force` will be supported. It is used by `edit-config` and I think `cordovaDependency` over-riding. 

I will have to look into `--searchpath=<directory>` & `--noregistry`

alexblom:
I can see once users are able to type a series of plugins/platforms in a more easy/familiar way, the batch install case may change. Historically they may just run 3-4 cordova plugin/platform adds, now it is more familiar to edit package.json, write the desired stack and run `cordova prepare`.

I have a feeling the dual listing of plugin names for variables is going to be error prone in this case (cordova-plugin-facebook4 from the example):

```
{
...
    "cordova": {
        "plugins": ["cordova-plugin-device", "cordova-plugin-facebook4"],
        //if the plugin has variables, store them
        "cordova-plugin-facebook4": {"APP_ID":"123456", "APP_NAME":"myApp"}
    }
...
}
```

Is there a particular reason the design is not closer to:

```
{
...
    "cordova": {
        "plugins": {
          "cordova-plugin-facebook4": {"APP_ID":"123456", "APP_NAME":"myApp"},
          "cordova-plugin-device": {}
    }
...
}
```

or something that doesn't require replicating the plugin name.

I need to think about this more and am unsure what I am proposing is much better yet. Just trying to avoid the case where a user changes workflows and now has an badly configured plugin.

purplecabbage:
We are stuck having at least one name dupe, but I do prefer your json frag @alexblom

stevengill:
@alexblom you make a good point. I chatted with @purplecabbage offline here and I agree it makes sense to avoid duplication where we can. Thinking of going with your proposed way for now for `cordova.plugins` but keeping `cordova.platforms` an array as none of the platforms take variables. 

```
    "cordova": {
        "plugins": {
          "cordova-plugin-facebook4": {"APP_ID":"123456", "APP_NAME":"myApp"},
          "cordova-plugin-device": {}
        },
        "platforms": ["ios"]
    },
   "dependencies" : {
         "cordova-plugin-facebook4" : "^1.1.8",
         "cordova-plugin-device" : "^1.4.1",
         "cordova-ios": "4.2.0"
    }
```

stevengill:
Updated proposal. Please review the PR! https://github.com/apache/cordova-lib/pull/499

PR #61, opened 2017-01-13
CB-12008 : autosave by default proposal for cordova@7

audreyso:
Proposal: Autosave will be the default option (when saving or removing dependencies) instead of using the --save flag.

### DIFF
--- a/proposals/autosave-by-default.md
+++ b/proposals/autosave-by-default.md
@@ -13,8 +13,13 @@ Autosave will be the default option.
 Example: cordova platform add ios
 This command will automatically save the ios platform.
 Include a --no-save flag to not save.
-Example: cordova platform add ios --no-save
+Example: cordova platform add ios --nosave
 This command will allow the ios platform to NOT be saved.
+
+### Notes
+Autosave will only be used in the cli right now (not cordova-lib).
+If you're using cordova-lib directly, you will still need to use the
+--save flag.
 
 ### Future Goals
 Include a global config to turn autosave off for cordova.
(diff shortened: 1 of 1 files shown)


stevengill:
I like the idea of switching to save by default for cordova@7

kerrishotts:
+1 to auto save for cordova@7. I'm consistently forgetting to add it to my cli commands even when I know about it. ;-)

remcohaszing:
Using `--save` screws up the layout of `config.xml`. Comments are removed, indentation is weird, blank lines are removed. I purposely don't use it.

As long as `--save` doesn't respect the existing `config.xml`, I vote against this.

stevengill:
Interesting point @remcohaszing. 

What type of information are you storing in the comments?

How is indentation weird? This is something we might be able to fix when we write out to config.xml. 

Most likely, we will take a similar stance to `package.json` and not have comments in config.xml.

audreyso:
Autosave will only be used in the cli right now (not cordova-lib).... you will still need to use the --save flag if using cordova-lib directly.

brodybits:
I raised a couple followup issues:
- [Apache CB-12397](https://issues.apache.org/jira/browse/CB-12397) - add .gitignore to standard app template
- [Apache CB-12398](https://issues.apache.org/jira/browse/CB-12398) - update users guide to document which application artifacts to include/exclude from git

Thanks @audreyso for your work to add the autosave by default.

audreyso:
Thank you for creating these issues! @brodybits

remcohaszing:
@stevengill 
Comments could be about anything. From my experience they're mainly about why certain versions are locked down. I am totally pro #53 though, which may seem ironic.

I just tried the `--save` flag again and the indentation isn't as bad as I remember.

PR #62, opened 2017-01-25
CB-11982 : proposal for creating a new cordova config command

audreyso:
CB-11982 : proposal for creating a new cordova config command


stevengill:
Looks like you added the autosave proposal into this PR by accident.

PR #63, opened 2017-01-25
CB-11982 proposal for creating a new cordova config command

audreyso:


stevengill:
I feel like this would land in `cordova-cli` repo and then pass the default values down to `cordova-lib`

PR #71, opened 2017-06-09
CB-12895proposal : jshint to eslint

audreyso:
Things to consider when transitioning from jshint to eslint...

### DIFF
--- a/proposals/jshint-to-eslint.md
+++ b/proposals/jshint-to-eslint.md
@@ -1,0 +1,68 @@
+# Jshint to Eslint Proposal
+- Status: Proposed
+- CB-12895
+
+## Current
+
+Currently, the .jschintrc files in most cordova repos have the following settings: 
+
+```
+{   "node": true
+  , "bitwise": true
+  , "undef": true
+  , "trailing": true
+  , "quotmark": true
+  , "indent": 4
+  , "unused": "vars"
+  , "latedef": "nofunc"
+  , "jasmine": true
+  , "esversion": 6
+}
+```
+
+## Overview
+
+This is a proposal to determine which settings/rules we should keep, ignore, or add when moving from jshint to eslint.
+
+When we switch to eslint, we will be able to keep the same jshint settings (listed above). 
+
+We are also using ‘extends’, which is taking a set of rules and adding them to your config. There are more than one 'extends' to pick from and so far we are using semi-standard because it supports checking for semi-colons. (Note : semi-standard adds Standard, and then adds the one rule on top of that for semicolons).
+
+When using semi-standard, eslint is pickier about a few more things than jshint. It has certain rules that are built in. These additional rules in eslint could either be just turned off and ignored if we don’t think they are necessary. (In some cases you can customize your rules. For instance, you can change the indenting from 2 to 4).
+
+Link to the standard rules —> https://standardjs.com/rules.html
+
+There are many rules and listed below are the ones that have come up most frequently when testing eslint using extending semi-standard in cordova-repos.
+
+1. Fixable: means that this rule is fixable using the --fix flag on the command line. The checks/rules below are all fixable.: 
+ 
+* `comma-spacing` - enforce spacing before and after comma
+* `keyword-spacing` - enforce spacing before and after keywords
+* `no-multi-spaces` - disallow use of multiple spaces
+* `space-before-function-paren` - require or disallow a space before function opening parenthesis
+* `spaced-comment` - require or disallow a space immediately following the // or /* in a comment
+* `space-infix-ops` - require spaces around operators
+ 
+2. Not fixable with --fix flag, but also come up a lot:
+
+* `brace-style` - enforce one true brace style
+* `camelcase` - require camel case names
+* `curly` - specify curly brace conventions for all control statements
+* `eqeqeq` - require the use of === and !==
+* `no-extra-boolean-cast` - disallow double-negation boolean casts in a boolean context
+* `no-throw-literal` - restrict what can be thrown as an exception
+* `operator-linebreak` - enforce operators to be placed before or after line breaks
+* `padded-blocks` - enforce padding within blocks
+
+## Proposal
+
+I propose that we keep the eslint semi-standard defaults in initially and remove padded-block, operator-linebreaker, and no-throw-literal.
+
+## Questions
+Would we like to use eslint's semi-standard default rules across the cordova repos or would we prefer to ignore certain ones and turn them off? Are there any other eslint rules not listed above that we would like to turn off or customize? Concerns? Thoughts? Pros and cons?
+
+## Links
+* Link to all the eslint rules —> http://eslint.org/docs/2.0.0/rules/
+* Link to the standard rules --> https://standardjs.com/rules.html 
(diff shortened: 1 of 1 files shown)


shazron:
I don't agree about having 'camelcase' in if its going to cause a lot of busywork changes in existing code for no benefit (perhaps a test run with it first).

I agree with Fil regarding regarding rolling across all the repos. It should be part of CI IMO (at least for the repos I've been using, they have been).

filmaj:
I'm using the linked-to eslintrc as I play around in the cordova-lib repo, and geez, there are a lot of violations!

Three specific rules I have hit so far:
 - at least temporarily, i need to explicitly set `'indent': ['error', 4]` to work with four-spaced projects.
 - agree with @shazron, forcing camelcase seems like busywork, I turned it off with `'camelcase': 'off'`.
 - at least in cordova-lib, we violate the `one-var` rule to not allow multiple `var` declarations split with a comma. That's usually paired with `no-multi-spaces` rule violation, as multiple `var` declarations split with commas are generally aligned along the `=` sign with multiple spaces. I actually like leaving these rules in place, though, so I would be in favour of keeping it and fixing the code up.

remcohaszing:
I think it's best to stick with an eslint preset without exceptions. This makes it easier to:

1. Maintain a simple and consistent configuration across repositories:

   ```yaml
   extends: standard
   ```

   I do see why it's useful to refactor code gradually, temporarily disabling rules for now.
2. This makes it simple to refer to a style guide, no exceptions.

audreyso:
@filmaj Thanks! Just wanted to confirm, you are in favor of keeping the no-multi-spaces and one-var rules ON so users CANNOT do:
var x, y, z;

audreyso:
Thanks for the feedback all! After looking at all input, here is what the .eslintrc.yml file could look like:

```root: true
extends: semistandard
plugins:
  - standard
  - promise
rules:
  indent:
    - error
    - 4
  quotes:
    - error
    - single
  camelcase: off
  no-undef: off
  padded-blocks: off
  operator-linebreak: off
  no-throw-literal: off
```
Just to sum it up:
- Indenting will require 4 spaces
- Quotes will require single quotes
- `camelcase` will be off
- `no-undef` will be off (I would like to add this one to the list b/c otherwise it will throw errors & want to define every “describe”, “it” , “expect”.)
- `padded-blocks` will be off
-  `operator-linebreak` will be off
- `no-throw-literal` will be off

We can use this consistency across repos.

dpogue:
Just throwing it out there for discussion (I haven't used it myself), but another option that seems popular is using [prettier](https://prettier.github.io/prettier/) to automatically format code at commit time.

filmaj:
@audreyso yes, that's my personal opinion, but I am open to whatever.

I would, however, leave `no-undef` on - it's super useful and gets rid of clutter. For the jasmine globals, we just have to make sure to set the env to jasmine. See here for an example: https://github.com/filmaj/cordova-lib/blob/no-lazy-load/spec-cordova/platform/index.spec.js#L17

remcohaszing:
Instead of setting the `eslint-env` comment per spec file, you could also specify it globally for all JS files. This may not be exactly correct, but it saves the overhead of setting it for every file.

Also you don't have to specify plugins if they're already defined by the extended config.

Also don't forget to specify `root: true`.

```yaml
root: true
extends: semistandard

rules:
  indent:
    - error
    - 4
  quotes:
    - error
    - single
  camelcase: off
  padded-blocks: off
  operator-linebreak: off
  no-throw-literal: off
```

Also, why is the `quotes` rule overridden? It appears that standard already enforces single quotes (https://github.com/feross/eslint-config-standard/blob/master/eslintrc.json#L153)

audreyso:
@remcohaszing Thanks for bringing this to my attention! I can add in root and also remove quotes/plugins.
@filmaj Thanks for sharing. I can also leave no-undef on and set the env to jasmine.


```root: true
extends: semistandard
rules:
  indent:
    - error
    - 4
  camelcase: off
  padded-blocks: off
  operator-linebreak: off
  no-throw-literal: off
```

audreyso:
hi everyone—> i’ve gotten different feedback about the eslint format and i was hoping we could choose one that works everyone! 
We have to decide how we want the following code to look for when we are using “return” and using “then”. There are 4 choices.

Currently it looks like this:
```module.exports.check_ant = function () {
    return superspawn.spawn('ant', ['-version'])
    .then(function (output) {
        // Parse Ant version from command output
        return /version ((?:\d+\.)+(?:\d+))/i.exec(output)[1];
    }).catch(function (err) {
        if (err) {
            throw new CordovaError('Failed to run `ant -version`. Make sure you have `ant` on your $PATH.');
        }
    });
};```

This is no longer acceptable with eslint will have to be changed.

Choice 1: move “then” so it’s in the same line as “return”
```module.exports.check_ant = function () {
    return superspawn.spawn('ant', ['-version']).then(function (output) {
        // Parse Ant version from command output
        return /version ((?:\d+\.)+(?:\d+))/i.exec(output)[1];
    }).catch(function (err) {
        if (err) {
            throw new CordovaError('Failed to run `ant -version`. Make sure you have `ant` on your $PATH.');
        }
    });
}; ```

Choice 2: leave “then” where it is and indent everything. (notice there will be a wider space now by the closing bracket)

```// Returns a promise. Called only by build and clean commands.
module.exports.check_ant = function () {
    return superspawn.spawn('ant', ['-version'])
        .then(function (output) {
            // Parse Ant version from command output
            return /version ((?:\d+\.)+(?:\d+))/i.exec(output)[1];
        }).catch(function (err) {
            if (err) {
                throw new CordovaError('Failed to run `ant -version`. Make sure you have `ant` on your $PATH.');
            }
        });
};```


Choice 3: disable individual lines lines so we can keep it as is

```.then(function (output) {  // eslint-disable-line ```

Choice 4:  take out the indenting as a rule.


and current config file looks like this:
```
root: true
extends: semistandard
rules:
  indent:
    - error
    - 4
  camelcase: off
  padded-blocks: off
  operator-linebreak: off
  no-throw-literal: off
```

If you have any suggestions about the rules such as which ones we should ignore and which are okay to keep (https://standardjs.com/rules.html) , please write them here! Thanks so much!

kerrishotts:
Personal preference would be option 2, solely because it keeps the first `.then` in roughly the same columns as later parts of the chain. With option 1, it's really easy to glide over the `.then`, at least for me. :-)

filmaj:
I have a preference for option 2, also OK with option 1. Do not like option
3.

What would option 4 look like in practice (more curious than anything)?

audreyso:
@filmaj for option 4:
you could either add ```/* eslint indent: 0 */```
to the individual file you want the indenting rule to ignore
OR 
(currently the default indenting rule is 4 spaces)
to turn off the indent rule for all files you could …
in the .eslintrc file you can disable indent:
rules:
  indent:
    - 0

filmaj:
4 is not an option for me, then.

stevengill:
I've already started using option 1 but like option 2 too. I agree option 2 is better for readability. I just dislike the closing bracket not being aligned with the original call. Either works for me. We should decide which one to go with though and start using that.

audreyso:
Okay great! Thanks everyone for your input...looks like option 2 is the preferred choice!

stevengill:
Closing since this has been implemented

PR #72, opened 2017-08-02
added cordova8 proposal

stevengill:
Some ideas we have for the next major version of cordova. Thoughts? 

Let me know what points need discussing in more detail.

### DIFF
--- a/proposals/cordova8.md
+++ b/proposals/cordova8.md
@@ -1,0 +1,53 @@
+We have come with a few ideas for the next major version of Cordova. 
+
+### Goals
+
+* Reduce maintenance burden of cordova tooling
+* Refactor cordova-lib 
+* Improve unit test coverage
+    
+### Proposed Changes
+
+* Remove non cordova-fetch fetching code
+    * remove no-fetch flag + any associated files with non cordova-fetch fetching
+    * drop npm dependency
+    * issue: https://issues.apache.org/jira/browse/CB-13055
+
+* Create new `cordova install` command
+    * remove restore functionality from `cordova prepare` command
+    * move restore functionality into new `cordova install` command. (installing plugins/platforms listed in `package.json` and `config.xml`)
+    * update docs to include `cordova install` 
+    * issue: https://issues.apache.org/jira/browse/CB-11981
+    * proposal: https://github.com/cordova/cordova-discuss/pull/54
+
+* Remove plugins dir from cordova projects
+This directory has traditionally been a staging directory. We copy plugins from the `plugins` directory to `platforms/platform/...` when a plugin needs to get installed. Now, these plugin directories also live in `node_modules` directory. I propose we delete the `plugins` directory and instead just copy the plugins straight from `node_modules`.
+    * TODO: investigate reason why `fetch.json`, + `<platform>.json` exist and if we need them. They live in plugins directory right now.
+    * https://issues.apache.org/jira/browse/CB-13059
+
+* Remove `cordova platform save` command
+This command isn't necessary anymore since we save by default since `cordova@7`. It was previously used to save added plugins and platforms to `config.xml`. 
+    * **Do Now** add deprecation notice for next `cordova@7` release.
+    * remove use of `cordovaProject/platforms/platforms.json`
+    * axe `cordova-lib/src/cordova.platform_metadata.js`
+    * issue: https://issues.apache.org/jira/browse/CB-13057
+
+* Remove deprecated platforms (non platform api compatibile platforms).
+This is just a continuation of what we started in `cordova@7`
+    * **Do Now** deprecate `webos` (in cordova@7)
+    * Remove `webos`, `bb10`, `ubuntu` + all associated files 
+    * issue: https://issues.apache.org/jira/browse/CB-13056
+    * previous `cordova@7` release
+        * deprecated `bb10` & `ubuntu` https://issues.apache.org/jira/browse/CB-12674
+        * removed polyfill support for previously deprecated platforms https://issues.apache.org/jira/browse/CB-11242
+        * removed `firefoxos` https://github.com/apache/cordova-lib/commit/41bc8dafa928855a690c70e0b950dc72f37ac9fc
+        * removed `wp8` https://github.com/apache/cordova-lib/commit/6f0a5611aadb56cb1bba8beeb3c22b45eeffc974
+        * removed `amazon-fireos` https://github.com/apache/cordova-lib/commit/0d637f2d67a95ef709443eaa9d7481eb14f79b8d
+
+* Incompatible plugins should make the build fail
+Currently, if a plugin fails the engine requirement it is skipped and the build continues. It makes more sense for the build to fail and have a force flag to make the build continue if that is the intended behavior. 
+    * issue: https://issues.apache.org/jira/browse/CB-12122
+
+* `plugin add` should use `cordovaDependency` in `package.json` instead of `engine` tag in `plugin.xml` when installing plugins
+The plan has always been to replace `engine` tag with `cordovaDependency`.
+    * issue: https://issues.apache.org/jira/browse/CB-13123


dpogue:
To the extent that we're using npm now for managing platforms and plugins, do we actually need a `cordova install` command or is `npm install` sufficient?  I guess the issue is the step of actually creating the platforms/[whatever] folders?

stevengill:
Removing platforms is a much bigger refactor I think. All the platform create scripts need to output the templates, cordova.js, other build artifacts to some location. We do directly call the create scripts from `node_modules/<platform>`, but those create scripts then output to the platforms directory. Platforms directory is essentially our dist/build directory. You can get access to generated Xcode/android studio projects. It could be something we look at down the line though. 

Just realized I should probably also add moving plugin/platform dependencies out of `config.xml` and have them only in `package.json`. Stop the auto syncing of `config.xml` and `package.json`. This currently happens during the restore step in `cordova prepare`.

dpogue:
Yeah, I'm not currently suggesting getting rid of the platforms folder, just wondering if it is created by `cordova install` or by `cordova prepare`? If it is created by `cordova prepare`, do we need `cordova install` at all?

filmaj:
I am in favour of all of these points. But oof, that seems like a whole lotta work. All for the better, I think!

@dpogue on the topic of `install` vs. `prepare`, I'd like to point out that stuffing more of this functionality into `prepare` is what spurred the discussion in the first place, as per [a comment on the `install` proposal](https://github.com/cordova/cordova-discuss/pull/54#issuecomment-250051369): "I agree that [restore] has been a horrible source of bugs, and still isn't quite right."

@stevengill [brought up a good point in that discussion](https://github.com/cordova/cordova-discuss/pull/54#issuecomment-250058857) that `install` feels like the right place to do project-setup type actions, rather than again stuffing into `prepare`.

Finally, I'd like to point to [my comment in that proposal](https://github.com/cordova/cordova-discuss/pull/54#issuecomment-316519057) to highlight that we should strive to keep the design of cordova-lib consistent. In order keep the `prepare` command true to its initial design, we should keep pre-compilation steps encapsulated in `prepare` (needed due to the abstracted-above-platform-specifics nature of cordova). As @stevengill pointed out, providing an API like `install` keeps the workflow or setup-specific tasks like pulling in dependencies and creating project configuration separate from `prepare`'s pre-compilation/build tasks.

To your question about whether we can get away with just running `npm install`: I think encapsulating install/post-install commands in a `cordova install` comamnd, and mirroring npm in that way, provides us a great stepping stone for eventually automatically running `cordova install` after an `npm install` (via npm scripts, for example). Worth exploring that more.

I would suggest any further discussion on the implementation details / specifics of `install` be debated in #54 instead of here. We can always bring the decisions / suggestions from that discussion back here (where `prepare` vs. `install` is only a subset of the larger changes being deliberated).

proposals/cordova8.md
```diff
@@ -0,0 +1,53 @@
+We have come with a few ideas for the next major version of Cordova. 
+
+### Goals
+
+* Reduce maintenance burden of cordova tooling
+* Refactor cordova-lib 
+* Improve unit test coverage
+
```
audreyso: Nice work on the proposal! I really like the idea of adding more unit tests! 👍

proposals/cordova8.md
```diff
@@ -0,0 +1,53 @@
...
+* Remove `cordova platform save` command
+This command isn't necessary anymore since we save by default since `cordova@7`. It was previously used to save added plugins and platforms to `config.xml`. 
+    * **Do Now** add deprecation notice for next `cordova@7` release.
+    * remove use of `cordovaProject/platforms/platforms.json`
+    * axe `cordova-lib/src/cordova.platform_metadata.js`
+    * issue: https://issues.apache.org/jira/browse/CB-13057
+
+* Remove deprecated platforms (non platform api compatibile platforms).
+This is just a continuation of what we started in `cordova@7`
+    * **Do Now** deprecate `webos` (in cordova@7)
```
audreyso: deprecation notice for webos
https://github.com/apache/cordova-lib/pull/585/files

proposals/cordova8.md
```diff
@@ -0,0 +1,53 @@
...
+    * proposal: https://github.com/cordova/cordova-discuss/pull/54
+
+* Remove plugins dir from cordova projects
+This directory has traditionally been a staging directory. We copy plugins from the `plugins` directory to `platforms/platform/...` when a plugin needs to get installed. Now, these plugin directories also live in `node_modules` directory. I propose we delete the `plugins` directory and instead just copy the plugins straight from `node_modules`.
+    * TODO: investigate reason why `fetch.json`, + `<platform>.json` exist and if we need them. They live in plugins directory right now.
+    * https://issues.apache.org/jira/browse/CB-13059
+
+* Remove `cordova platform save` command
+This command isn't necessary anymore since we save by default since `cordova@7`. It was previously used to save added plugins and platforms to `config.xml`. 
+    * **Do Now** add deprecation notice for next `cordova@7` release.
```
audreyso: deprecation notice for platform save
https://github.com/apache/cordova-lib/pull/585/files

proposals/cordova8.md
```diff
@@ -0,0 +1,53 @@
...
+    * remove use of `cordovaProject/platforms/platforms.json`
+    * axe `cordova-lib/src/cordova.platform_metadata.js`
+    * issue: https://issues.apache.org/jira/browse/CB-13057
```
audreyso: started the work for removing platform save 13057
https://github.com/apache/cordova-lib/pull/586

proposals/cordova8.md
```diff
@@ -0,0 +1,53 @@
...
+
+* Remove deprecated platforms (non platform api compatibile platforms).
+This is just a continuation of what we started in `cordova@7`
+    * **Do Now** deprecate `webos` (in cordova@7)
+    * Remove `webos`, `bb10`, `ubuntu` + all associated files 
+    * issue: https://issues.apache.org/jira/browse/CB-13056
```
audreyso: started 13056 --> https://github.com/apache/cordova-lib/pull/588

proposals/cordova8.md
```diff
@@ -0,0 +1,53 @@
...
+    * issue: https://issues.apache.org/jira/browse/CB-11981
+    * proposal: https://github.com/cordova/cordova-discuss/pull/54
+
+* Remove plugins dir from cordova projects
+This directory has traditionally been a staging directory. We copy plugins from the `plugins` directory to `platforms/platform/...` when a plugin needs to get installed. Now, these plugin directories also live in `node_modules` directory. I propose we delete the `plugins` directory and instead just copy the plugins straight from `node_modules`.
+    * TODO: investigate reason why `fetch.json`, + `<platform>.json` exist and if we need them. They live in plugins directory right now.
+    * https://issues.apache.org/jira/browse/CB-13059
+
+* Remove `cordova platform save` command
+This command isn't necessary anymore since we save by default since `cordova@7`. It was previously used to save added plugins and platforms to `config.xml`.
```
shazron: Would users upgrading from a project using an older cordova version still need this, to save their existing platforms and plugins?
If we remove this, they will have to manually add them in config.xml, no?

stevengill: Looking at the analytics (assuming I'm doing this correctly), the command has been used 64 times out of the past 1.9 Million requests. Less that 0.1 percent. Pretty clear sign that it is not being used much.

I do see the usefulness of the command for users pre 7 who didn't used to use `--save` (which has been around much longer). I just don't think people in that group are even aware of the platform save command.

I think by this point asking users to modify config.xml or readd plugin + platforms is fair. They will probably have to readd/update older platforms or plugins anyways. 

Removing it will clean up cordova-lib a bit and allow us to get rid of these extra generated config files that go into peoples cordova projects

filmaj: Thanks for digging into the analytics @stevengill. It is clearly not a use case we need to worry about much, which is good, simplifying the path forward!

brodybits:
Should we merge, close, or do something else with this one?

PR #73, opened 2017-08-07
Better integrate with module bundlers

remcohaszing:
I'd love to see better integration with module bundlers, such as webpack.

I think I figured out the global steps of what needs to be done for this.

I might be missing a lot of details, because I'm not that familiar with Cordova internals.

### DIFF
--- a/proposals/manual-bundling.md
+++ b/proposals/manual-bundling.md
@@ -1,0 +1,150 @@
+# Allow developers to bundle Cordova JS files manually
+
+## Goal
+
+Make Cordova integrate better with a modern JavaScript setup.
+
+Basically it comes down to that Cordova should allow using node modules. This allows better integration with tooling such as:
+
+- Package managers ([npm] / [yarn])
+- Module bundlers ([webpack] / [browserify] / [rollup])
+- IDEs (Atom ([autocomplete-modules], [js-hyperclick]))
+- Linters ([eslint-plugin-import])
+- Miscellaneous tooling ([depcheck])
+
+
+## Current situation
+
+Let's say I would like to build a Cordova app for Android using the camera plugin.
+
+Steps:
+
+1. Create a Cordova project using `cordova create`.
+2. Remove the legacy hooks directory and obsolete .npmignore.
+3. Add the desired Android platform and camera plugin.
+
+   Now this is where things get weird.
+
+   A directory `/platforms/android/platform_www` is created.
+
+   Inside this folder is a file `cordova.js`, which contains a UMD bundle with the following modules:
+
+   - `'cordova'` → https://github.com/apache/cordova-js/blob/master/src/cordova.js
+   - `'cordova/exec'` → https://github.com/apache/cordova-android/blob/master/cordova-js-src/exec.js
+   - `'cordova/platform'` → https://github.com/apache/cordova-android/blob/master/cordova-js-src/platform.js
+   - `'cordova/android/*'` → https://github.com/apache/cordova-android/tree/master/cordova-js-src/android
+   - `'cordova/plugin/android/*'` → https://github.com/apache/cordova-android/tree/master/cordova-js-src/plugin/android
+   - `'cordova/*'` → https://github.com/apache/cordova-js/tree/master/src/common (except for exec and platform, not sure if they're the `_b` postfixed ones, but it doesn't really matter)
+
+   In addition there is a `cordova_plugins.js`, which maps the plugins to a file location.
+
+   Also there is a `plugins/cordova-plugin-camera/www/*`, which contains the files from https://github.com/apache/cordova-plugin-camera/tree/master/www, but with a `.` as a path separator, whereas in any other situation this would be a `/`.
+
+   However, if I take a look at what the `cordova` package on npm is, it's the CLI, which is totally unrelated to the UMD bundle.
+   
+   In addition to `/platforms`, there's also a `/plugins` directory. This contains the following:
+   
+   - Copies of the plugins that already exist in `node_modules`, but with the dotfiles missing.
+   - `android.json` — This appears to duplicate some info from `package.json` and `config.xml`, but in another format.
+   - `fetch.json` — This appears to duplicate some info from `package.json`, but in yet another format.
+
+
+## Proposal
+
+As a modern JavaScript developer, I would like to use *<this year's popular tool>* to bundle my dependencies and code.
+
+- Get rid of `cordova.js`.
+- Get rid of `www/`.
+- Get rid of `plugins/`.
+
+### 1. Load plugins from the `node_modules` directory, not `plugins`
+
+Cordova already downloads packages to `node_modules` using `cordova-fetch`. There's no need to maintain a copy of them in the `plugins` directory.
+
+This also allows plugin creators to use npm dependencies for both hooks and browser code the way this would be done for other packages.
+
+### 2. Let a developer require Cordova using its own build tools.
+
+Based on [cordova-plugin-camera] as an example:
+
+As a developer, I'd like to turn this:
+
+```js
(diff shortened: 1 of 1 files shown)


proposals/manual-bundling.md
```diff
@@ -0,0 +1,150 @@
+# Allow developers to bundle Cordova JS files manually
+
+## Goal
+
+Make Cordova integrate better with a modern JavaScript setup.
+
+Basically it comes down to that Cordova should allow using node modules. This allows better integration with tooling such as:
```
stevengill: yes! This would be sweet

proposals/manual-bundling.md
```diff
@@ -0,0 +1,150 @@
...
+
+## Proposal
+
+As a modern JavaScript developer, I would like to use *<this year's popular tool>* to bundle my dependencies and code.
+
+- Get rid of `cordova.js`.
+- Get rid of `www/`.
+- Get rid of `plugins/`.
+
+### 1. Load plugins from the `node_modules` directory, not `plugins`
```
stevengill: An issue for this has already been filed https://issues.apache.org/jira/browse/CB-13059. Also it is proposed in the cordova@8 proposal https://github.com/cordova/cordova-discuss/pull/72

proposals/manual-bundling.md
```diff
@@ -0,0 +1,150 @@
...
+
+Cordova already downloads packages to `node_modules` using `cordova-fetch`. There's no need to maintain a copy of them in the `plugins` directory.
+
+This also allows plugin creators to use npm dependencies for both hooks and browser code the way this would be done for other packages.
+
+### 2. Let a developer require Cordova using its own build tools.
```
stevengill: So if i'm understanding this correctly, you are suggesting we use html imports to do plugin loading. 

Few concerns:

1) I think this would be a tough sell for plugins. Pretty much every plugin would have to change. 

2) Our goal is to mimic web apis. Camera is a bad example because it isn't based on a standard. But a lot of the current work on our plugin audit is about deprecating non standard apis (We are keeping camera don't worry) and updating apis of plugins where the standard has changed since we implemented it originally. 

3) imports aren't supported everywhere i think. Would that force users to use a build tool?

proposals/manual-bundling.md
```diff
@@ -0,0 +1,150 @@
...
+
+```js
+// CORDOVA_PLATFORM is set by the build tool, e.g. using webpack.DefinePlugin.
+require(CORDOVA_PLATFORM)
+```
+
+Plugins would make calls to `cordova-js/exec` (replacing `cordova/exec`) independant of the platform. If a developer wishes to use an outdated plugin, it's the developer's responsibility to alias `cordova/*` calls to `cordova-js/*`.
+
+
+### 4. Dependency management
```
stevengill: We looked into peer dependencies at one point. We found out they were getting deprecated. https://docs.npmjs.com/files/package.json#peerdependencies

They don't auto install peer deps anymore. They just omit warnings to the user. 

The reason why cordova needs to manage plugin deps is because we can't allow a plugin to have two different versions installed in one project. 

For example, say file-transfer plugin requires file version 3. And say media plugin requires file version 4. This would and should cause an issue. Since the native implementations of file would clash on the native side (duplication errors). 

npm has never given us a way to manage this unique case so we have always had to have our own level of dep management

proposals/manual-bundling.md
```diff
@@ -0,0 +1,150 @@
...
+
+Cordova doesn't need to define special dependency management for plugins. `npm` can handle this.
+
+Cordova plugins and platforms depend on the same instance of `cordova-js`. For this reason all Cordova platforms and plugins should specify `cordova-js` in their `package.json`'s `peerDependencies`.
+
+Also if plugins depend on other plugins, this can be handled by specifying them as `peerDependencies`.
+
+
+### 5. Fix up the browserify setup
```
stevengill: the browserify code is confusing. With the refactoring of cordova.js we could potentially remove this and allow our devs to use one of the various build step tools in their own setup

proposals/manual-bundling.md
```diff
@@ -0,0 +1,150 @@
...
+CordovaCamera.getPicture(cameraSuccess, cameraError, cameraOptions);
+```
+
+This can be done by adding the following to `package.json`:
+
+```json
+  "main": "www/Camera.js"
+```
+
+### 3. Use [cordova-js] as a proxy to the platform
```
stevengill: cordova-js definitely needs to be rethought out. This is an interesting idea I'd like to hear more discussion about.

dpogue:
As someone who uses Cordova to build web apps, I'm pretty opposed to anything that requires my app code to know that Cordova exists.  I'm building apps for the web, Cordova is just a packaging tool and polyfill loader (in the form of plugins). My app should not know or care that it's running in Cordova.

Currently Cordova requires the app to include `<script src="cordova.js"></script>` in the index page, which is slightly annoying but easy enough to workaround with a empty cordova.js file for web.

Plugins can already require node modules using npm via their package.json files if Cordova is run with the `--browserify` option (which I believe is default now?)

stevengill:
@dpogue `browserify` is not default.

proposals/manual-bundling.md
remcohaszing: > So if i'm understanding this correctly, you are suggesting we use html imports to do plugin loading.

I'm not suggesting to use html imports, but I'm suggesting to require a build tool. These tools are already specialized at handling node modules, so Cordova doesn't have to handle this differently. Anywhere where I write "`import`", it might just as well say "`require`".

> Few concerns:

> - I think this would be a tough sell for plugins. Pretty much every plugin would have to change.

One implementation of such a build hook could mimic the old behaviour for a while.

- Copy the contents of `wwww` into the platform www directory.
- Use browserify (or webpack) to detect all Cordova plugins and the Cordova platform directory, bundle those and write this into `cordova.js` in the platform www directory.
- For backwards compatibility, this could map `cordova` imports to `cordova-js` and `cordova/*` to `cordova-js/*`.

This could be included by default in `cordova-app-hello-world`.

I would personally prefer a hook that would bundle and minify everything, including my own code, my own dependnecies and Cordova code. This means I would have to write my own hook for this.


> - Our goal is to mimic web apis. Camera is a bad example because it isn't based on a standard. But a lot of the current work on our plugin audit is about deprecating non standard apis (We are keeping camera don't worry) and updating apis of plugins where the standard has changed since we implemented it originally.

Mimicking (or polyfilling) web APIs doesn't require any special handling from Cordova. Just assign values to the `window` object.


> - imports aren't supported everywhere i think. Would that force users to use a build tool?

This would force users to include a build hook. This is likely to be based on a build tool.

proposals/manual-bundling.md
remcohaszing: > We looked into peer dependencies at one point. We found out they were getting deprecated. https://docs.npmjs.com/files/package.json#peerdependencies

> They don't auto install peer deps anymore. They just omit warnings to the user.

> The reason why cordova needs to manage plugin deps is because we can't allow a plugin to have two different versions installed in one project.

> For example, say file-transfer plugin requires file version 3. And say media plugin requires file version 4. This would and should cause an issue. Since the native implementations of file would clash on the native side (duplication errors).

> npm has never given us a way to manage this unique case so we have always had to have our own level of dep management


Their behaviour has changed between npm versions 2 and 3, but it's not deprecated as far as I know.

I believe giving a warning **is** the way npm has given to manage flat dependencies.

- ESLint recommends it for shareable configs: http://eslint.org/docs/developer-guide/shareable-configs#publishing-a-shareable-config
- React libraries add `react` as a peer dependency.
- Angular even uses peer dependencies for depending on `@angular/*` packages.

I don't think the situation for Cordova is any different. How npm should handle these missing dependencies is out of scope for Cordova. Users shouldn't be surprised if their builds fail when using mismatched versions.

PR #75, opened 2017-08-08
GitHub migration proposal

audreyso:
Proposal to move issues from Jira to Github


kerrishotts:
+1 to moving to GitHub Issues. Option 1 makes the most sense to me, personally.

proposals/GithubMigration.md
```diff
@@ -0,0 +1,58 @@
+# Migration of Jira Issues to Github
+- Status: Proposed
+
+## Overview
+This is a proposal to migrate Jira issues over to GitHub. [See issue CB-13157](https://issues.apache.org/jira/browse/CB-13157)
+
+## Current
+Currently, issues for all Cordova repos are tracked on [Jira](https://issues.apache.org/jira/secure/Dashboard.jspa).
```
shazron: I would use this URL issues.apache.org/jira/browse/CB

proposals/GithubMigration.md
```diff
@@ -0,0 +1,58 @@
...
+This proposal recommends that issues are mirgated from Jira to Github so that the issues are with the code. This is recommended because:
+* Jira can be a barrier to actually creating issues.
+* This is an expected open source workflow.
+* This helps to modernize Cordova.
+* Issues should be with the code.
+
+There are two potential options when using Github to track issues:
+
+## Option 1
+When issues are migrated to Github, each repository will contain its own set of issues. In each repo, there will be a list of the issues, which can be found under the Github issue tab. Therefore, the issues can be found with the code. This can allow developers to reference pull requests more easily. Developers will also be able to "watch" repos and receive notifications when pull requests are sent or when new issues are created. We encourage and want to make it easy for committers to watch Github for the repos they want to be involved in.
```
shazron: - Github "Issues" tab

proposals/GithubMigration.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Option 2
+When issues are migrated to Github, they are added to one "issues" repository. There will be one master “issues” repo that can track all of the Cordova issues on GitHub.
```
shazron: Not sure about this one :P

infil00p:
I agree with Option 1. The only downside is linking issues, but you know, people could just include the link to their issue or PR in the issue thread they're participating in.

Also, each maintainer should probably write a rationale to how they triage issues, for example, only the following things are blockers for me on Cordova-Android:

1. Hello World Doesn't Run
2. Cordova doesn't Compile
3. The Device Plugin doesn't install (or Echo, or some plugin that's representative of the absolute minimum functionality of plugin installation that we can't blame on the plugin itself)

BTW: We will probably still have to use Private JIRA for Security Issues, since we can't discuss those publicly until the issue is resolved as per the Apache Security Policy.

http://www.apache.org/security/committers.html

purplecabbage:
Personally, I think issues are better left in JIRA.

Our current system does a good job of flagging JIRA Issue IDs and maintaining that link.
Distributing our issues to me, means more issues, in more places, and more chance to miss things.

Any plugin issue is quite likely to apply to 1 or more platforms, and possibly even tooling.  If we move to gh-issues, this means we expect people who file issues to figure out where all the repos are to file the issue, then which of those repos is most appropriate, ...

We need to handle security issues very carefully, I have yet to see how we will approach this.

PS. Also, a great example of gh workflow, is the simple fact that this discussion is only half of the story, and there is an equally relevant dev-list thread [1] happening. 

[1] https://lists.apache.org/thread.html/a992eaf62c8475a4576784ca3f9d8acc575220f0c53e6016bc4455e7@%3Cdev.cordova.apache.org%3E

janpio:
I suggest that `cordova-docs` could be one of the first or the first "components" to be migrated. It has a small number of issues in JIRA, has some people actively working on translations and fixing stuff in the source content right now, which would mean that issues would used quite quickly. 

Docs are also a prime target for contributions by volunteers in general, so this would really benefit from the easier workflow on Github.

(It would also enable tighter integration with the Crowdin workflow per their Github integration!)

stevengill:
Thanks for sharing that link @purplecabbage. I somehow missed that whole conversation.

I'm hugely in favor of switching to github issues. I've always ran into people who don't know where to file issues or won't file issues because they have to create a new account. 

I prefer the github workflow of having issues live with the code + prs. 

Plugin issues get buried in Jira. Unless someone is actively working on a plugin, the issues only get looked at during release time. From my experience, I'd say that open PRs get looked at way more than open issues. It would be way easier for me to go to a plugin repo and see how many open issues exist than go and search jira. In terms of what platforms are affected, I think we could potentially use platform labels/tags. 

I made a comment on security issues in the email thread but I'll share it here too.

_For security, we already get requests sent to our private mailing list (private@cordova.apache.org). We then create a private issue to track the problem and only make it public once we solve the problem and do a release._ 

_Instead of a private jira instance, we could use a private github repo. Follow the same steps otherwise._

Alternatively, for security we could keep using private jira.

filmaj:
> Our current system does a good job of flagging JIRA Issue IDs and maintaining that link. Distributing our issues to me, means more issues, in more places, and more chance to miss things.

@purplecabbage while it works, I don't think it does a good job. Already we experience having discussions in multiple places between JIRA and GitHub PR discussions (and as you mentioned, this proposal suffers from the same!). It works, but is noisy, hard to parse, and confusing. I think consolidating issues and PRs into one place in GitHub would clean that up tremendously.

> Any plugin issue is quite likely to apply to 1 or more platforms, and possibly even tooling. If we move to gh-issues, this means we expect people who file issues to figure out where all the repos are to file the issue, then which of those repos is most appropriate, ...

We already experience this in JIRA-land. It is up to contributors and committers to do triage work, determine if an issue is valid, clean up issue description, tag the appropriate component and label, etc.

I'll post what I noted in the dev list thread:

> I think the ease of use of GitHub issues overcomes potential problems about cross-referencing issues. Worth noting on this topic that GitHub already provides good support for referencing pull requests from issues across repos / orgs. The benefit of having issues and PRs in one place, to me, is too tasty to pass up.

proposals/GithubMigration.md
```diff
@@ -0,0 +1,58 @@
...
+* Triaged
+* Priority (low, medium, high)
+* Waiting for user/waiting for contributor
+* In progress
+* Easy fix
+* Backlog
+* Duplicate
+* Milestones (Github feature)
+
+## Additional Work
```
filmaj: I think we should also update the GitHub pull request templates. They reference JIRA in them, so I think the suggested steps therein won't be relevant depending on the outcome of this discussion.

Android has an existing JIRA issue filed to tackle this: https://issues.apache.org/jira/browse/CB-12848
So does Browser: https://issues.apache.org/jira/browse/CB-13037

filmaj:
Oh and one more thing: Apache JIRA regularly craps out for me. Several times a day requests to the Apache JIRA instance time out (like right now!). I have built up the muscle memory to write up issue reports in my local text editor as more than once I've lost those during these timeout scenarios.

proposals/GithubMigration.md
janpio: Additionally, you should think about creating an Issue template per repo. Some examples I am aware of:

https://raw.githubusercontent.com/ionic-team/ionic-cli/master/.github/ISSUE_TEMPLATE.md
https://raw.githubusercontent.com/ionic-team/ionic-app-scripts/master/.github/ISSUE_TEMPLATE.md
https://raw.githubusercontent.com/ionic-team/ionic/master/.github/ISSUE_TEMPLATE.md
https://raw.githubusercontent.com/fastlane/fastlane/master/.github/ISSUE_TEMPLATE.md

(It might make sense to not rush this though, to see what actually happens and what has to be "directed" to make more sense and create better issues)

purplecabbage:
Okay, I am now convinced this is a good idea, especially with the recent frequent jira outages.

Can we be systematic in our approach?
Maybe pick a few key repos/components to start with and try it out? The tooling (cli/lib/serve/fetch) might be a good place to start, although I doubt most users would know which of those modules their error was occuring ... maybe we should start with platforms. docs would be good too, as @janpio suggested.

janpio:
An idea on the "users can't find the correct repo" problem: 
Create a repo "issues" (or even more general "repositories") that has all features disabled (Issues, Projects, Wiki, PRs) and only contains a README with a list of the repositories (probably platforms and cli first, then docs and tooling, last libs) and which functionality or problems probably belong to it. Set that as the new redirect target of issues.cordova.io. (Can also link to instructions to report security problems.)

proposals/GithubMigration.md
stevengill: @audreyso lets add these recommendations to this section of the PR. 

Update pull request templates (link to relevant issues)
Add issue template

stevengill:
Okay, so I think next steps for this is to 

1) get infra to enable issues on repos that have moved over to gitbox (android, ios, browser, windows)
2) start migrating issues we want to move over. I imagine this will mean creating a duplicate of the issue on github, adding the original jira issue link to github issue, linking the new github issue back to the old jira issue, close the old jira issue. 
3) Move forward on additional work section of this PR

janpio:
Has there been any thoughts on _how_ to actually migrate issues? 

Relevant questions:
What user will post (and then "host") these issues?
What information (JIRA fields) will be transferred?
How will comments or changes be represented?
Manually or by script?
Any default labels for "transferred" issues?

janpio:
(Was curious and googled a bit about migration scripts, some relevant links:
https://gist.github.com/jonmagic/5282384165e0f86ef105
https://github.com/doctrine/jira-github-issues + https://github.com/doctrine/jira-github-issues/issues/2
https://gist.github.com/jonmagic/5282384165e0f86ef105)

stevengill:
Thanks @janpio! I haven't spent time looking for migration scripts yet but will review what you sent.

In terms of your question, if it was a manual process, the idea was to copy over relevant comments. For labels, try to mimic what labels the issue already has. Also, follow the label suggestions from this PR to add new/missing labels. 

As for which issues get migrated, it would be up to the discretion of the migrator. Rough guidelines would include:
* how old is the issue, how active
* what is the priority of the issue
* is it something worth copying over.
* only open issues 

Of course if we use a migration script, this would be much simpler and faster :)

janpio:
(One more though: A mapping from known JIRA users to known Github usernames might be beneficial for notifications of replies - for all the Cordova team people a mapping could be created to enable importing with correct account usage)

brodybits:
Is this idea still under consideration? Can someone list exactly what is still blocking this idea?

I think people would be much more likely to report the issues they encounter on GitHub than on the JIRA system.

brodybits:
Bump

brodybits:
I already started <https://github.com/brodybits/cordova-notes> as a temporary workaround ref: #87

brodybits:
Bump again

shazron:
We can start with one repo at first (thus one jira component), see how it goes (like cordova-docs like @janpio mentioned). But, if new issues are filed on JIRA, we need a way to filter/detect this, and either automatically respond to the user to file in Github, or we manually reply ourselves for redirect the user to the new repo (may get tedious). 

We can do an `all or nothing` by doing a migration script, and only push the most recent (365 days?) issues, and setting JIRA to read-only, with a warning note in the JIRA site header (if we could, we have to ask INFRA)

janpio:
(`cordova-windows` already has Github issues activated, no action taken in JIRA when this happened)

brodybits:
I would still love it if we could try this for cordova-docs. I still think this is badly needed.

shazron:
Filed https://issues.apache.org/jira/browse/INFRA-16503

janpio:
@brodybits To move this forward "someone" would have to define how to handle JIRA issues, create the necessary infrastructure (migration script etc.) for it, test it and create a proposal to execute that on the dev mailing list.

shazron:
@brodybits the INFRA issue is resolved. Check the cordova-docs repo.

brodybits:
Thanks @shazron!

@janpio I think @shazron gave a nice suggestion how to do this in <https://github.com/apache/cordova-discuss/pull/75#issuecomment-387326561>.

I wonder if we should consider a more brute force approach where we close all existing documentation issues with a message to report in <https://github.com/apache/cordova-docs/issues>, and hopefully we can set JIRA to close any new Cordova doc issues with the same kind of message.

stevengill:
IMO, Just declare bankruptcy on jira. Comment on every open issue that we moved to github and close them. Refile if it is still an issue. Make jira readonly.

Step 1: enable issues on all repos
Step 2: update site to point to new issue trackers. Update repos themselves to do the same. PR templates need updates I think. Also need to add issue templates. 
Step 3: find script to comment on all open issues, message should say we are moving over to github issue tracker (link to list of all issue trackers). Ask them to re file over there if the problem still exists. Close issue 
Step 4: make jira read only

Thoughts?

purplecabbage:
JIRA is required for security issues though isn’t it?

stevengill:
@purplecabbage yeah, looks like GitHub hasn't solved this yet. 

How about we keep JIRA open, but only for security issues. Any new issues coming in get closed if they aren't security related. 

I've updated my comment above to reflect this

dpogue:
Are there any tools available to us to do things like automated moving of issues from one repo to another? I'm quite concerned that we'll get issues for plugins reported to platform repos, or issues for everything reported to things like cordova-lib or cordova-cli. I don't want it to turn into a full-time job to triage these issues and be constantly redirecting people to re-open them on other repos. If it involves too much time commitment from anyone, it's just going to end up not getting triaged and then we'll have the same situation as JIRA but with even less ability to search and filter.

stevengill:
maybe the issue template in every repo links to the full list of issue templates and tells users to double check they are filing it in the right one?

Menardi:
@dpogue We could use [Google's Github Issue Mover](https://github.com/google/github-issue-mover) or the Chrome extension [https://chrome.google.com/webstore/detail/kamino/ffdebockfdjileaojbbccofhgncmioaf](Kamino) for moving issues. It's definitely not an ideal workflow, but it doesn't seem like Github are planning on implementing this any time soon, as per https://github.com/github/hub/issues/886.

On the topic of automating the issue process, I think we should enable labelling as part of the initial reporting process. By default, Github only lets a user add a label after the issue is created, which they may easily forget to do. It would be possible to have a bot which would look for checkboxes in the issue, and apply the relevant label automatically. The Apollo project has [their own bot](https://github.com/apollographql/apollo-bot) which does something like this. Would it make sense to have a Cordova issue bot which could do this (and possibly more)?

shazron:
We should put the proposal up for a vote on dev@cordova.apache.org with all the proposal details that Steve put out, plus any amendments in the discussion. 

I like the bot proposal -- we can explore if Apache has some way for us to host a bot.

Menardi:
Since the `cordova-docs` repository has already been moved to GitHub, I have created a pull request to update the documentation to reflect this. I also created an issue template and updated the pull request template, based on this discussion. Any feedback would be appreciated!

https://github.com/apache/cordova-docs/pull/823

purplecabbage:
This was done ...

PR #105, opened 2018-07-24
`cordova platform add` refactor

dpogue:
Proposal for splitting the behaviour of `cordova platform add` into several smaller steps that could be reused by other parts of the tooling and more easily tested.

See also: https://github.com/apache/cordova-common/pull/34 for a first pass at some of the lower-level package.json support needed for this.

/fyi @raphinesse @erisu @brodybits


proposals/cordova-platform-add-refactor.md
```diff
@@ -0,0 +1,160 @@
...
+        platform_module = 'cordova-' + platform_spec
+    }
+
+    // Get the platform name with any version specifiers stripped off
+    let platform_name = platform_spec.slice(0, platform_spec.lastIndexOf('@'))
+
+    let pkgJson = new PackageHelper('package.json')
+
+    // TODO: See if there's an easy way to determine if the platform is already
+    // installed with a matching version and avoid the npm invocation
```
raphinesse: cordova-fetch already does this, so you can just call it and use the path to the installed package it returns to determine the package name too.

dpogue: Ah nice!

PR #107, opened 2018-08-06
CB-10071: (ios) Proposal swift support

knight9999:
The purpose of this proposal is to support iOS plugin written by Swift.
I propose introducing two functions. 

- The first function will edit the `Bridging-Header.h` according to the settings defined in `plugin.xml`. 

- The second is setting the application’s Swift version according to the settings defined in `config.xml`.

Note that we can specify the Swift version for the application. Plugins with different Swift version can not be imported simultaneously. This is contrary with CocoaPod libraries in which Swift version can be specified for each.

### DIFF
--- a/proposals/ios-swift-plugin-support.md
+++ b/proposals/ios-swift-plugin-support.md
@@ -1,0 +1,70 @@
+# Swift Plugin Support
+- Status: Proposed
+
+## Overview
+
+The purpose of this proposal is to support iOS plugin written by Swift.
+I propose introducing two functions. 
+
+- The first function will edit the `Bridging-Header.h` according to the settings defined in `plugin.xml`. 
+
+- The second is setting the application’s Swift version according to the settings defined in `config.xml`.
+
+Note that we can specify the Swift version for the application. Plugins with different Swift version can not be imported simultaneously. This is contrary with CocoaPod libraries in which Swift version can be specified for each.
+
+
+## Bridging-Header
+
+Introduce the `BridgingHeader` type attribute value in the `header-file` tag in plugin.xml
+
+ex.
+```
+<header-file src="src/ios/Hoge-Bridging-Header.h" type="BridgingHeader" />
+```
+
+This specification is similar to the proposal https://issues.apache.org/jira/browse/CB-10071 where the value is `SwiftObjcBridgingHeader`. The specification here is little simpler.
+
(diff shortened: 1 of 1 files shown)


dpogue:
This seems good, and should remove the need for people to fiddle with hook scripts or extra plugins just to get Swift working.

Regarding plugins written in Swift, how do you envision they would specify the compatible Swift version? I guess they could inject it into the platform's config.xml and let the app's config.xml override it?

Plugin specifies Swift 4.0 in its plugin.xml:
```xml
<!-- plugin.xml -->
<plugin ...>
  <platform name="ios">
    <config-file parent="/*" target="config.xml">
      <preference name="SwiftVersion" value="4.0" />
    </config-file>
  </platform>
</plugin>
```

Application config.xml specifies Swift 4.1:
```xml
<!-- config.xml -->
<widget ...>
  <platform name="ios">
    <preference name="SwiftVersion" value="4.1." />
  </platform>
</widget>
```

Result: Swift 4.1

shazron:
Sounds good. Are you planning to do this for cordova-ios@5, the next major version?

dpogue:
I think we should try to get this in for cordova-ios@5

knight9999:
I have found one difficulty.

Assuming that there are two plugins, `pluginA` and `pluginB`.

- `pluginA` is written in pure Obj-C plugin with the header file `pluginA.h`.

`pluginB` is written in pure Swift but depends on pluginA.

pluginB's methods need to call onto pluginA's methods.

In this use case, our specification is not enough.

To solve this issue, pluginB would be required to have an additional obj-c header file

*pluginB.h*
```
#import 'pluginA.h'
```

and in pluginB's `plugin.xml` file, specify this file as `BridgingHeader`.
```
<header-file src="src/ios/pluginB.h" type="BridgingHeader" />
```

dpogue:
That still seems acceptable. It makes sense that pluginB would be required to define its dependency on pluginA, even if the process for doing that involves an extra header file.

knight9999:
I have just sent two PRs.

PR for cordova-ios
https://github.com/apache/cordova-ios/pull/395
This cordova-ios PR will require the following cordova-common PR to be merged first.

PR for cordova-common
https://github.com/apache/cordova-common/pull/42

knight9999:
I have just sent a PR for cordova-doc.

https://github.com/apache/cordova-docs/pull/880

PR #110, opened 2019-09-03
UIWebView Strategy for Cordova iOS

dpogue:
Background: As of August 2019, Apple is now showing a deprecation warning when uploading apps to the App Store that include UIWebView-related code. As a result, all Cordova apps built for iOS receive this deprecation warning on upload.

We need to determine how Cordova as a project wants to proceed to solve this problem. There are pull requests already being opened that I'm quite concerned with end up breaking all existing Cordova apps, and I think we want to be more careful about that.

I've put together this document outlining some options for discussion.

### DIFF
--- a/proposals/UIWebView-strategy.md
+++ b/proposals/UIWebView-strategy.md
@@ -1,0 +1,86 @@
+UIWebView Strategy for Cordova iOS
+==================================
+
+As of August 2019, Apple is now showing a deprecation warning when uploading apps to the App Store that include UIWebView-related code. UIWebView has been unofficially deprecated in favour of the new WKWebView for a while, but Apple has now made it officially deprecated. As a result, all Cordova apps built for iOS will receive this deprecation warning on upload. Currently it is just a warning, but Apple has not provided a timeline for when it will start rejecting apps that use UIWebView APIs.
+
+We've had [a plan](https://cordova.apache.org/news/2018/08/01/future-cordova-ios-webview.html) for several years to make it easier to migrate to WKWebView and eventually deprecate UIWebView use by Cordova, but almost no progress has been made on this objective since the plan was outlined over a year ago.
+
+Cordova users are clamoring for a fix to the problem and there are pull requests open to remove UIWebView entirely, but if we proceed without a concrete plan then we are guaranteed to cause a lot of pain and suffering for Cordova users.
+
+The goal of this document is to outline several potential migration plans, and facilitate discussion on the pros and cons of each.
+
+
+High-Level Goals
+----------------
+
+* Make WKWebView the default web view for Cordova iOS applications
+* Make it possible to build Cordova iOS apps without including any UIWebView APIs
+* Maintain compatibility as much as possible with existing plugins
+* Provide a means for Cordova apps to transition from UIWebView to WKWebView
+
+
+Challenges
+----------
+
+* UIWebView-related classes are potentially part of the public API exposed to plugins
+* Moving all apps over to WKWebView will result in data being lost (localStorage, indexedDB, etc.)
+* WKWebView does not support all the same features as UIWebView and has stricter security requirements, and **many existing Cordova iOS apps will not work if run in WKWebView** without changes
+* The current WKWebView plugin needs updates to use Scheme Handlers to avoid issues that require a local web server.
+* The InAppBrowser plugin also uses UIWebView. Moving to WKWebView would potentially lose some functionality.
+    * Ideally, the InAppBrowser should be deprecated in favour of SFSafariViewController which severely restricts functionality in the name of security. Currently InAppBrowser use could largely be considered bad practice.
+* Few contributors with very limited time, and significant work to be done to implement any of these strategies
+
+
+Cordova-iOS Migration Options
+-----------------------------
+
+### 1. Cordova-iOS 6 drops UIWebView entirely
+
+This is probably the easiest option, but is also a very large breaking change that potentially breaks existing plugins and existing apps.
+
+The steps are roughly:
+
+1. Move the existing WKWebView plugin into cordova-ios.
+2. Update to use Scheme Handlers with WKWebView.
+3. Remove all the existing UIWebView code.
+
+
+### 2. Cordova-iOS 6 defaults to WKWebView, with option to disable UIWebView
+
+This is a more gradual transition, closer to what had originally been announced, but runs the risk of Apple setting a rejection deadline that's more aggressive than what we're able to meet.
+
+The steps are roughly:
+
+1. Move the existing WKWebView plugin into cordova-ios, with a preference to pick the webview engine.
+2. Update to use Scheme Handlers with WKWebView.
+3. Update the "Hello Cordova" template to have the webview preference default to WKWebView.
+4. Add warnings to cordova-ios for apps that are using the UIWebView preference.
+5. Wrap all the UIWebView code behind a compiler definition that can be controlled by a preference, to allow disabling all the UIWebView code.
+6. [Next Major] Disable the UIWebView code by default, but allow it to be enabled by the preference for apps that require it.
+
+
+InAppBrowser Migration Options
+------------------------------
+
+### 1. Drop UIWebView code entirely
+
+As above, this is easiest, but a potentially breaking change for existing apps.
+
+### 2. Pick webview based on preference
+
+Similar to above, the webview would be determined based on a config.xml preference. The UIWebView code would also need to be conditionally guarded by a compiler definition.
+
+Steps would be:
+
+1. Update IAB plugin to pick the webview based on the config.xml preference.
+2. Wrap the UIWebView code in compiler definitions that are controlled by a preference, matching cordova-ios.
+
+
+### 3. Deprecated IAB
+
+Probably also a major breaking change, but this would be better in the long term.
+
+An ideal scenario would be to integrate SFSafariViewController into Cordova-iOS (rather than as a plugin) and allow any URL that is not in the allow-navigation list to open in SFSafariViewController automatically. Apps would lose the ability to customize the page, but that has been a major security concern for years (particularly around OAuth login pages).
+
+> Out of scope, but if this is the decided strategy, Cordova-Android should also be updated to use Chrome Custom Tabs in the same situations, so that IAB can be deprecated for both platforms at the same time.
+


NiklasMerz:
I personally would vote for **option 2** in case of cordova-ios and inappbrowser. It looks like this is the best compromise and backwards compatible way. App developers should have moved to WKWebView with the plugin for some time and this way the upgrade should be the easiest way.

We are still relying on IAB for some legacy sites embedded in the "new" app in Cordova. There are plans to get rid of them in the future but that may take a long time and could cause compatibility issues with old versions of our product.

That IAB is a bad practice should be made clear with advice how  to use it, but I would argue that this is not a strong enough reason to drop it completely. But I must admit that SafariViewControler looks like the way to go in the far future.

NiklasMerz (commented):
Add this file to README?

proposals/UIWebView-strategy.md
```diff
@@ -0,0 +1,86 @@
...
+
+
+Challenges
+----------
+
+* UIWebView-related classes are potentially part of the public API exposed to plugins
+* Moving all apps over to WKWebView will result in data being lost (localStorage, indexedDB, etc.)
+* WKWebView does not support all the same features as UIWebView and has stricter security requirements, and **many existing Cordova iOS apps will not work if run in WKWebView** without changes
+* The current WKWebView plugin needs updates to use Scheme Handlers to avoid issues that require a local web server.
+* The InAppBrowser plugin also uses UIWebView. Moving to WKWebView would potentially lose some functionality.
```
janpio: InAppBrowser has supported WKWebview for quite some time via `usewkwebview`.

proposals/UIWebView-strategy.md
```diff
@@ -0,0 +1,86 @@
...
+    * Ideally, the InAppBrowser should be deprecated in favour of SFSafariViewController which severely restricts functionality in the name of security. Currently InAppBrowser use could largely be considered bad practice.
```
janpio: Very much different use case and functionality, also not related to UIWebView<->WKWebView. This should not be part of this discussion.

What probably should be part, is how the default InAppBrowser would break when UIWebView is removed from `cordova-ios`.

What should also be part of the discussion, is that IAP should probably do the same for dropping UIWebView and switch to WKWebView as default.

proposals/UIWebView-strategy.md
```diff
@@ -0,0 +1,86 @@
...
+Cordova-iOS Migration Options
+-----------------------------
+
+### 1. Cordova-iOS 6 drops UIWebView entirely
+
+This is probably the easiest option, but is also a very large breaking change that potentially breaks existing plugins and existing apps.
+
+The steps are roughly:
+
+1. Move the existing WKWebView plugin into cordova-ios.
```
janpio: Interesting to note: There are multiple, different WKWebView plugins out in the ecosystem. Just integrating the Cordova one would probably break all those.

janpio:
I think https://github.com/ionic-team/cordova-plugin-ionic-webview is the elephant in the room - how does this very popular plugin fit in here?

https://npm-stat.com/charts.html?package=cordova-plugin-ionic-webview&package=cordova-plugin-wkwebview-engine&from=2018-09-03&to=2019-09-03

proposals/UIWebView-strategy.md
NiklasMerz: IAB with WKWebView basically works fine now except this issue https://github.com/apache/cordova-plugin-inappbrowser/issues/492

IAB also needs to hide UIWebView behind a build flag, like some other plugins.

NiklasMerz:
> I think https://github.com/ionic-team/cordova-plugin-ionic-webview is the elephant in the room - how does this very popular plugin fit in here?
> 
> https://npm-stat.com/charts.html?package=cordova-plugin-ionic-webview&package=cordova-plugin-wkwebview-engine&from=2018-09-03&to=2019-09-03

Like said in the document the apache wkwebview engine needs to implement Scheme handlers. Scheme handlers in the ionic plugin make CORS easier, thats probably why it is the more popular choice apart being the default in new Ionic apps.

janpio:
Possibly true, but not my point.

Should/Could this much more popular plugin be integrated? How can this plugin keep existing after Cordova iOS switched over to WKWebView (all those current users of it need a migration path)?

Ionic should be _an active part_ of this discussion.

NiklasMerz:
Good point. @jcesarmobile is probably the right person to ask.

jcesarmobile:
Bringing a previous discussion link where we agreed to rename the cordova-plugin-wkwebview-engine classes on cordova-ios integration so it doesn't break existing WKWebView plugins that started as a fork (probably all of them), and keep the plugable webviews so people can still use different plugins from the default. 
https://markmail.org/message/fxe6q6nnernvr6h6.

The problem with Scheme handlers is they are iOS 11+, so we should deprecate iOS 10. We can probably fallback to file urls on iOS 10, but I think having different behaviours is worse for the users.

We have 2 PRs at the moment:
[first one](https://github.com/apache/cordova-ios/pull/662) add a preference that enables the compiler definition to disable UIWebView if desired. I think this is a third option, similar to the option 2, but as long as we don't disable it by default, we can release this as a minor release as it's a new feature and not a breaking change. With this in place it should be easier to transition to option 2 in a major version where we change the default from UIWebView to WKWebView. This also "unlock" people that are not using UIWebView but still get the warning.

The [second PR](https://github.com/apache/cordova-ios/pull/663) claims that the previous one didn't work for him and still got the warning. This one is for 1, removing UIWebView entirely, but doesn't rename the classes, so other plugins will collide and fail to install. Also doesn't have url scheme handlers as it's a direct port of current plugin.

So I personally like more the first one, but we should check if we still get the warning and try to figure out why to properly fix it. 

About integrating the Ionic one, it has a little Ionic Appflow code that I don't think fits into cordova-ios, so better keep them separate. But I can give a hand with the scheme handlers as I implemented it on the ionic one. But as I said, that would mean deprecating iOS 10.

timbru31:
> I think ionic-team/cordova-plugin-ionic-webview is the elephant in the room - how does this very popular plugin fit in here?

> Ionic should be an active part of this discussion.

I totally agree with you Jan!  
How about the folks over at @ionic-team give something back to the Cordova community by contributing their cordova-plugin-ionic-webview fork back to the (orignal) cordova-plugin-wkwebview-engine since the fork already has been migrated to handle, e.g., the scheme handler for iOS 11+? IMHO this would help the Cordova project to tackle this big hurdle much faster.


---

Regarding the proposed options from Darryl, I would favor the "hard removal" options. We are very limited in regards of the development capacity and removing old, legacy, and deprecated code instead of maintaining both of them makes things easier.

NiklasMerz:
Just another thought. I don't know if it makes sense or is technically possible.

Move UIWebView to a plugin as well and make cordova-ios use this or some of the WKWebView plugins. cordova-ios just contains the code to load a webview. By default the next version should load the Apache  WKWebViewEngine. Users who need UIWebView, the Ionic one, a fork or others can specify their own one like it is now. 

That way no UIWebView would be in cordova-ios and it would still be flexible enough like today.

jcesarmobile:
That could work too, but it's a breaking change as they will need to install a webview plugin, and probably more work to move the UIWebView code into a separate plugin that will be short lived.

proposals/UIWebView-strategy.md
dpogue: > Very much different use case and functionality, also not related to UIWebView<->WKWebView. This should not be part of this discussion.

Agreed, but I also feel that deprecating IAB is in line with "the goal of [Cordova] is to cease to exist" now that there are platform-supported better options.
I'll drop this point from the discussion.

dpogue:
> The problem with Scheme handlers is they are iOS 11+, so we should deprecate iOS 10.

This seems fine to me, but should probably be brought to the list for discussion. iOS 13 is rolling out soon, and Apple traditionally only supports the two most recent iOS versions.

> we can release this as a minor release as it's a new feature and not a breaking change.

I feel pretty strongly that we should not be making any of these changes in a minor version bump.

> keep the plugable webviews so people can still use different plugins from the default.

We will definitely be keeping the pluggable webview support, so it should be possible for people to keep using other webview plugins.

> we agreed to rename the cordova-plugin-wkwebview-engine classes on cordova-ios integration so it doesn't break existing WKWebView plugins that started as a fork

Thank you for reminding me of this. I'll update the document to make it clear that we need to rename classes when we integrate.

jcesarmobile:
> This seems fine to me, but should probably be brought to the list for discussion. iOS 13 is rolling out soon, and Apple traditionally only supports the two most recent iOS versions.

We unofficially also only support the last 2 versions, but we have been slow at removing them and if we do it fast people complain.

> I feel pretty strongly that we should not be making any of these changes in a minor version bump.

If you see the first PR is pretty "simple". If you don't have the preference it will behave the same way it behaves at the moment, if you have the preference it will switch a few lines in compile time. I think that can safely go into a minor release and give us more time for a complete integration in next major release.

> We will definitely be keeping the pluggable webview support, so it should be possible for people to keep using other webview plugins.
> Thank you for reminding me of this. I'll update the document to make it clear that we need to rename classes when we integrate.

Related to the first point too. As we are integrating the WKWebView plugin renaming the classes, adding the scheme handlers and keeping the plugable webviews, in worst case, people can manually change deployment target to iOS 10 and still use the current cordova-plugin-wkwebview-engine that uses file instead of the scheme handlers.

proposals/UIWebView-strategy.md
jcesarmobile: It's definitely a better option for what IAB was originally created, safe web browsing.
But as security is its strong suit, it's not as feature rich as IAB, all code/css injection, events, etc. that people relies on nowadays is not possible, so we can't deprecate it.

NiklasMerz: I agree. I have to maintain some pages which rely on that for some time. If you want your pages in IAB tightly integrated to your app you need this script injection stuff. Deprecating it should be the last resort, since it works fine now if you know what you are doing and use it in a responsible way. So security is not really a point.

bpresles:
After reading this discussion, I found a good idea to rename classes of embedded WKWebView engine so that they don't conflict with existing engine plugins.

So the PR 663 (https://github.com/apache/cordova-ios/pull/663) includes that change.

Also I created a `UIWebView` plugin based on current default `UIWebView` engine in `cordova-ios` (https://github.com/bpresles/cordova-plugin-uiwebview-engine) and published a first release : https://www.npmjs.com/package/cordova-plugin-uiwebview-engine
The funny thing is that this `UIWebView` engine can be used even on current and previous releases of `cordova-ios` (even though it's plain useless ;) ) as I took care to rename the classes ;)

bryyyanribo:
@bpresles, What do you think is the best approach to this problem? 

Should I Install this Cordova [plugin](https://github.com/ionic-team/cordova-plugin-ionic-webview)?

proposals/UIWebView-strategy.md
```diff
@@ -0,0 +1,86 @@
...
+
+The steps are roughly:
+
+1. Move the existing WKWebView plugin into cordova-ios, with a preference to pick the webview engine.
+2. Update to use Scheme Handlers with WKWebView.
+3. Update the "Hello Cordova" template to have the webview preference default to WKWebView.
+4. Add warnings to cordova-ios for apps that are using the UIWebView preference.
+5. Wrap all the UIWebView code behind a compiler definition that can be controlled by a preference, to allow disabling all the UIWebView code.
+6. [Next Major] Disable the UIWebView code by default, but allow it to be enabled by the preference for apps that require it.
+
```
NiklasMerz: I would add the "Move UIWebView" to plugin option to this document, too. Here a first draft based on comments on this PR. 

```suggestion

### 3 . Move UIWebView to a plugin and remove it from Cordova-ios
Move UIWebView to a plugin as well and make cordova-ios use this or some of the WKWebView plugins. cordova-ios just contains the code to load a webview. By default the next version of Cordova-ios should load the Apache WKWebViewEngine (Plugin installed by default on new apps, Migration instruction for old apps). Users who need UIWebView, the Ionic one, a fork or others can specify their own one like it is now.

There is already a proof-of-concept version of this done: https://github.com/bpresles/cordova-plugin-uiwebview-engine with https://github.com/apache/cordova-discuss/pull/110#issuecomment-528088573

This is a breaking change since existing apps need to install that plugin and creates work for a plugin that is probably short lived. InAppBrowser needs some work to get this working.
```

bpresles:
> @bpresles, What do you think is the best approach to this problem?
> 
> Should I Install this Cordova [plugin](https://github.com/ionic-team/cordova-plugin-ionic-webview) for me to temporarily fix the issue regarding UIWebView?

Adding a `WKWebView` engine plugin won't remove the warning with current `cordova-ios` version as `cordova-ios` still embedded the `UIWebView` based engine even if you install a `WKWebView` engine plugin` that replaces it at execution time.

But yes, that's a good idea that you install a `WKWebView` engine plugin in your project to prepare it for the future, because you may (will?) have to adapt some part of your code to make it compatible with `WKWebView`, and so doing it the sooner possible makes sure that your app code is already prepared for when Apple will reject apps with `UIWebView` and so for future version of `cordova-ios` that will stop using `UIWebView`.

manish2535:
Any updates on this guys?

jcesarmobile:
On the removal of the UIWebView, we should also remove the CDVUserAgent class, or modify it for the WKWebView.
Right now it uses a "hack" to change the UIWebView user agent, but that doesn't work on WKWebView, WKWebView has a customAgent property that can be used instead.

proposals/UIWebView-strategy.md
purplecabbage: the 'code/css injection' was always controversial, I was always against it as I wanted it to be a separate sandbox.
Pushing this off to wkwv is probably fine, and making a firm recommendation of SFSafariViewController when apps just need static pages, and/or oauth.

ragcsalo:
What about this plugin?
https://github.com/apache/cordova-plugin-wkwebview-engine

Harmonickey:
@ragcsalo  read the comments above, I think that should answer your question, they don't believe using this plugin will remove the warning message since it still embeds UIWebView

"Adding a WKWebView engine plugin won't remove the warning with current cordova-ios version as it still embed the UIWebView based engine even if you install a WKWebView engine plugin` that replaces it at execution time."

ragcsalo:
Indeed... :-/ I added the plugin, and removed the inappbrowser plugin, but still getting the error. Looking forward to having a solution soon enough. :-)

purplecabbage:
@ragcsalo re: ' but still getting the error'
Which error?

ragcsalo:
ITMS-90809: Deprecated API Usage - Apple will stop accepting submissions of apps that use UIWebView APIs . See https://developer.apple.com/documentation/uikit/uiwebview for more information.

breautek:
> I mean the deprecation warning.
> 
> ITMS-90809: Deprecated API Usage - Apple will stop accepting submissions of apps that use UIWebView APIs . See https://developer.apple.com/documentation/uikit/uiwebview for more information.

There is no solution at this moment. That is precisely what this ticket is for, discussing how Cordova should solve this solution.

proposals/UIWebView-strategy.md
janimakinen: What about LocalStorage from HTML5? That stuff is used all over the place at least by many apps I've made with Cordova. We can always make a backend system to send XMLHttpRequests to store data, but LocalStorage is just so much faster for small apps.

If I want to help make LocalStorage a possibility, would cordova-ios be the project to try to contribute to, or try to figure out solutions here?

NiklasMerz: I am not sure if LocalStorage should be part of this discussion. LocalStorage should work in WKWebView just fine. The only problem is that switching from UIWebView to WKWebView makes your apps data inaccessible with the new webview. There are some solutions out there to migrate. But this is not a new problem since WKWebView has been around some time. 

I don't know if this migration should be part of cordova-ios.

a3den:
We have the same problem on 6 Cordova based apps.
We installed WKWebView plugin and fix the inAppBrowser call in order to use WKWebView but Apple rejected yet.
Please be quick,

Perh0rd:
@a3den  Apple do not reject app for now, it's just a warning.

a3den:
@Perh0rd Yes, but that Apple email scares our customers...

proposals/UIWebView-strategy.md
alextkd: They can also use SafariViewController for inAppBrowser

talvespe:
@a3den 
There are tons of cordova based apps, everyone are scared but stay calm.
At this moment is just a warning from Apple. 

_keep following_
https://github.com/dpogue/cordova-discuss/blob/uiwebview-plan/proposals/UIWebView-strategy.md

Perh0rd:
I don't know if it's interested or not. But we use Cordova, and I never had any warning from Apple? Is that because we are using WkWebView plugin?

NiklasMerz:
Interestingly we never got that warning, too. We are using the WKWebView, InAppBrowser and lots of plugins but the last updates didn't get that warning about the deprecated API.

rrath-edr:
What is the latest here. I have multiple apps and I don't want to loose our client. Please this has been 3 months now.

Perh0rd:
People still using the default Cordova UIWebview , you should definitely first , think to migrate to wkwebview using the https://github.com/ionic-team/cordova-plugin-ionic-webview. It will be a good start and will bring only positive points to your app.

Ios 13 is really not optimise for UIWebview. Its a lot of work , because you will have to make some kind of adjustements.

But  if you want to prepare for the future of your app using Cordova , I think you have no choice.

We are not affiliated to any organization than our , and I only here give my personnal point of view after , we migrated part of our apps to wkwebview.

rrath-edr:
With new ionic versions by default it comes with ionic-webview and which says they are using WKWebview. But every time I do a search in Xcode I see the UIWebview keyword and apple sends me warning saying UIWebview has been deprecated and apple will reject the app. If I check in ionic issue tracker they redirect to cordova.

dafi:
On 9 October I've published on App Store an ionic 4 app and got no warnings, the version target on XCode is iOS 11 maybe the warning appear when the target is iOS 13?

rajashekaranugu:
> an ionic 4 app and got no warnings, the version target on XCode is iOS 11 maybe does the warning appear when the target is iOS 13?

Hi Dafi, This sounds like a relief, Would be great if can you also please share your ionic info details, Below is what i am currently using, Thanks in Advance !!! 

"$ ionic info
Ionic:

   Ionic CLI                     : 5.4.4 (/usr/local/lib/node_modules/ionic)
   Ionic Framework               : @ionic/angular 4.8.1
   @angular-devkit/build-angular : 0.801.3
   @angular-devkit/schematics    : 8.1.3
   @angular/cli                  : 8.1.3
   @ionic/angular-toolkit        : 2.0.0

Cordova:

   Cordova CLI       : 9.0.0 (cordova-lib@9.0.1)
   Cordova Platforms : android 8.1.0, ios 5.0.1
   Cordova Plugins   : cordova-plugin-ionic-keyboard 2.2.0, cordova-plugin-ionic-webview 4.1.1, (and 17 other plugins)

Utility:

   cordova-res : 0.8.0 
   native-run  : not installed

System:

   ios-deploy : 1.9.4
   ios-sim    : 8.0.2
   NodeJS     : v10.16.3 (/usr/local/bin/node)
   npm        : 6.4.1
   OS         : macOS Mojave
   Xcode      : Xcode 11.1 Build version 11A1027"

breautek:
Here is my thoughts, now that I've been pondering over it for awhile... I think `cordova-ios@6` should just simply drop `UIWebView` altogether using Option 1. Here is why...

It sounds quicker and easier. And the faster we can drop `UIWebView` the faster companies can migrate their apps and the faster they can have peace of mind that they don't need to scramble to update and deploy their apps before the currently undisclosed apple deadline. It being easier sounds like less opportunities for bugs as well. I realise developers can already migrate to `WKWebView` via the existing plugins, but as already mentioned, the core of `cordova-ios` still contains references to `UIWebView`, so migrating to any of the existing plugins is not really enough.

We don't know the exact date, but we do know that no matter what people is going to have to migrate away from `UIWebView`, so continuing to support `UIWebView` pass version 5 seems like a waste of resources. If any developer wishes to remain on `UIWebView` for whatever reason, they can remain on on version 5 until they are forced to migrate by Apple.

If they are any plugins that will be affected by this drop, then they declare support for up to cordova-ios 5, and have a major release for >= cordova-ios 6. Each of these plugins can be documented in a cordova-ios 5 to 6 migration guide.

I also don't think cordova-ios should include migration between `UIWebView` local storage and `WKWebView` local storage, but I do think this is a legit concern. This isn't a new problem and I'm pretty sure there are plugins available that does this migration and I think making a note in a migration guide on this will be sufficient.

To recap, I see no value in supporting deprecated feature, especially one that Apple will eventually block app deployments for beyond version 5. People who are stuck on version 5 because they need  `UIWebView` will have to migrate eventually one way or another. And these people will run into the same problem even if Cordova 6 did support `UIWebView` behind a flag. Why make things more complicated? I can't vouch for this PR because I don't really have the knowledge for native iOS development, but there is a PR that I believe does Option 1. https://github.com/apache/cordova-ios/pull/663

brodybits:
I would also really favor option 1, in the next major release asap. Supporting old and deprecated APIs has risk of breakage at an unpredictable point in the future, diverts time and attention from other fixes and improvements badly needed. Major release is always the perfect time to break things that need to be broken, such as old and deprecated APIs.

jcesarmobile:
Quicker is the conditional flag like the proposed [here](https://github.com/apache/cordova-ios/pull/662). It could go in a minor release since by default it behaves the same way as it does now. That unlocks people that is not using UIWebView anymore but get the warning because of cordova-ios. 

But for 6.x I agree, UIWebView should be removed. There is no point on keeping it.

brodybits:
> Quicker is the conditional flag like the proposed [here](https://github.com/apache/cordova-ios/pull/662).

apache/cordova-ios#662

Any ideas how we can move forward with this?

Do any members have time to review and publish this update?

MZap:
We have a lot of suggestions, comments and more in this PR. Are we stuck in analysis paralysis? Does Ionic have a path forward?

jcesarmobile:
To be clear, I mean like that PR, not that PR as that one only works if you have the setting before adding the platform or if you remove and add again. Ideally it should change the settings by running `cordova prepare`. Not 100% sure if it's possible.

breautek:
> Quicker is the conditional flag like the proposed [here](https://github.com/apache/cordova-ios/pull/662). It could go in a minor release since by default it behaves the same way as it does now. That unlocks people that is not using UIWebView anymore but get the warning because of cordova-ios.
> 
> But for 6.x I agree, UIWebView should be removed. There is no point on keeping it.

It still kinda feels more like a band-aid to me. It could be a temporary solution if there was a concrete deadline that we were close to missing, but I think we are still early enough to not worry about that. If that PR would to be merged, it would be something that would be ideally removed in 6.x as the flag would become obsolete. Therefore I think it just makes more sense to focus more on a 6.x release with the `UIWebView` drop, and same people who is already on `WkWebView` can just upgrade to 6.x.

brodybits:
> It still kinda feels more like a band-aid to me.

Agreed but the "band-aid" solution proposed in apache/cordova-ios#662 seems to be cleaner than apache/cordova-ios#663.

> I think it just makes more sense to focus more on a 6.x release with the `UIWebView` drop

If we want to focus on the new release, I would favor integrating apache/cordova-ios#662 with the following kind of replacements:

- `s/if WK_WEB_VIEW_ONLY/if 1/`
- `s/if !WK_WEB_VIEW_ONLY/if 0/`

jcesarmobile:
yeah, it's a band-aid

just saying the whole removal of the UIWebView, WKWebView integration, using WKSchemeHandlers and all other proposed changes are not an easy and quick fix and will take a lot of time before we can have a 6.0.0. #663 only covers a part of it and it's already a massive change hard to review.

In the meantime, people are scared because of the warning with no permanent solution, no matter how much times you tell them it's just a warning since there is no official date from Apple.

breautek:
> just saying the whole removal of the UIWebView, WKWebView integration, using WKSchemeHandlers and all other proposed changes are not an easy and quick fix and will take a lot of time before we can have a 6.0.0.

Ahh ok, I was under the impression that option 1 was the easiest option, as that appears to be perhaps an early consensus.

I think most of us agree that `UIWebView` should be completely dropped in 6.x, but if that is still a ways out, then I think https://github.com/apache/cordova-ios/pull/662 is a good alternative to get into a "safe" position quickly, as you suggested.

ragcsalo:
So there is still no quick solution? We're doomed..... :/ Any news on the deadline?

brunnobb:
> So there is still no quick solution? We're doomed..... :/ Any news on the deadline?

**Not even deadline on the news yet. Hold on. in any urgent case there is the ionic capacitator but I much prefer cordova**

Anyway. I agree that in the 6.0 losing local storage its a price to pay.
A good way to avoid is to use a file based approach using cordova-plugin-file.
There is the option to update the app now, migrating from HTML APIs to file APIs and stop relying on the browser.
Then nothing is lost when the changes happen.

breautek:
>Anyway. I agree that in the 6.0 losing local storage its a price to pay.
A good way to avoid is to use a file based approach using cordova-plugin-file.
There is the option to update the app now, migrating from HTML APIs to file APIs and stop relying on the browser.

This is exactly what my team has decided to do. Not only we are running into the same problems with local storage cause of webview changes on iOS, our android apps are still on crosswalk, and the latest version of crosswalk runs on chrome 53, so no css support to handle notches/cutout displays. 

> So there is still no quick solution? We're doomed..... :/ Any news on the deadline?

https://github.com/apache/cordova-ios/pull/715 has landed in master, but there is no npm release yet. If you want to try it out now, you can try `cordova-ios@nightly`, or fork the `cordova-ios` repo. This PR is to address the issue for version 5.x. Note that with this flag enabled you are expected to have a `wkwebview` plugin installed.

alextkd:
Anyone, any clue when an official release version will contain these changes?  5.1.0 doesn't seem to

jcesarmobile:
which changes?
5.1.0 will have the conditional compile time flag that disables UIWebView
6.0.0 will remove the whole UIWebView as that's a breaking change, will take a few months as there are a lot more changes needed that have not been done yet

ragcsalo:
Once there is a solution, please make/write a short tutorial on what we should do to migrate from UIWebKit to WKWebKit. Thanks! :-)

jcesarmobile:
What 5.1.0 does is to avoid the warning for people that has already migrated and are not using UIWebView, so if you have not migrated, 5.1.0 won't do anything and you'll just get a white screen if you enable the new preference but don't have a `WKWebView` plugin installed.

Some of the WKWebView plugins are:
- [the official apache plugin](https://github.com/apache/cordova-plugin-wkwebview-engine)
- [ionic plugin](https://github.com/ionic-team/cordova-plugin-ionic-webview)

I don't think there is a migration guide and I don't think there will be, but major pain points are:
- XHR/fetch calls to servers are affected by CORS, so you need to configure your server/end points
- Data loss. Since it's a different webview the webview data is lost (localStorage, etc). There are a few plugins that can help with data migration. Data in native side is not lost (file, sqlite, storage, etc. plugins that write data in native places)

Probably more, but those are the two I remember.

rajashekaranugu:
> * the official apache plugin

@jcesarmobile, thanks for the info. As you said Am also facing the issue regarding the data Loss. 

In our current app in production we were using cordova 4.5.0 and ionic-webview 1.2.1 and saving was some data to local storage.

So with this solution (using 5.1.0) am building the app with updated ionic-webView to 4.1.3 and the data in local storage is lost.

can you please direct me to those few plugins you referred in your second point in regards to Data Loss.

jcesarmobile:
[This one](https://github.com/benjamineberle/cordova-plugin-migrate-localstorage) is from UIWebView to ionic webview plugin, but I've not tried it.
Since you are in ionic webview plugin already that won't work, not sure if there is some plugin to migrate from v1 to v4, sorry.

[This one](https://github.com/MaKleSoft/cordova-plugin-migrate-localstorage) is to migrate to the apache one

rajashekaranugu:
@jcesarmobile, Thanks for the response. 
I have found https://github.com/kassamina/cordova-plugin-ionic-migrate-storage. which is used to migrate local storage from ionic-web-view v1 to v4. I know its off the topic, but might help for some one who is looking for this issue.

ASHBAW:
How do you verify that you have eliminated all instances of UIWebView?

My Ionic 4 project use the following plugins

```
 "cordova-ios": "^5.1.1",
 "cordova-plugin-ionic-webview": "^4.1.3",
 "cordova-plugin-inappbrowser": "git+https://github.com/kleeb/cordova-plugin-inappbrowser.git"
```

I have also added 
`<preference name="WKWebViewOnly" value="true" /> `
to my config.xml

I build from from the command line and then open of the project in Xcode. If I search for UIWebView I still see instances in a number of .m files.

If I unzip the generated .ipa file and grep through (grep -lr "UIWebView" *) the following is returned: MyApp.app/MyApp

So I am not quite sure what else I need to do?

Any insight would be greatly appreciated.

breautek:
@ASHBAW 

Do note that the `WKWebViewOnly` preference only applies to the `cordova-ios` package and does not apply to plugins. (Unless if they have been updated to read and respond to that preference of course)

I'm unfamiliar with that fork but `inappbrowser` is a plugin that still references `UIWebView`.

beevelop:
@breautek Yep, we experienced the very same issue with the inappbrowser plugin. I tried tackling it with PR https://github.com/apache/cordova-plugin-inappbrowser/pull/584 already.

The compile-flag solution might be quite useful for other plugin maintainers too. Is there a suitable place in the Cordova docs to note this down?

breautek:
> Is there a suitable place in the Cordova docs to note this down?

The flag should be added to https://github.com/apache/cordova-docs/blob/master/www/docs/en/dev/config_ref/index.md#preference imo. Not sure if there is an existing PR for this.

ASHBAW:
The fork came from this thread
[apache/cordova-plugin-inappbrowser/issues/583]
where @kleeb removed UIWebView code for the inappbrowser plugin
Without it my build failed as described in 583

ASHBAW:
Just a quick question. I use Ionic 4 which uses the cordova-plugin-ionic-webview by default.

I performed the following steps:

1. Created an empty started project 
2. Installed cordova-ios@5.1.1
3. Add <preference name="WKWebViewOnly" value="true" /> to my config.xml file.
4. Ran cordova prepare ios to apply the changes
5. Built a production version on the command line
6. Opened up project in Xcode

If I search the Xcode project I still see 117 references to UIWebView in 20 different files. 

Should I still be seeing references to UIWebView at this point?

If I build and .ipa file and unzip it (i.e. change .ipa to .zip and unzip) followed by doing a grep through the Payload folder it returns that the app has references to UIWebView:

>> grep -lr "UIWebView" *
>> MyApp.app/MyApp

I am thinking I should not have any references at this point?

jcesarmobile:
Yes, it won't remove the code, so you'll see the references if you search in Xcode. But as it's behind a build flag, Apple won't complain about it.

ASHBAW:
Got it thank you.

l3ender:
> The App Store will no longer accept new apps using UIWebView as of April 2020 and app updates using UIWebView as of December 2020.

https://developer.apple.com/news/?id=12232019b

orbitaloop:
Hi, so I wonder what's the official Cordova solution to this? 
thanks

jcesarmobile:
Solution to what? Did you read the thread?

ragcsalo:
I also haven't seen the solution yet in the thread. Using Cordova with Android Studio (and not ionic). Will there be a new Cordova version sometimes that would solve this, or do we have to use the WKWebView plugin and re-write our code completely to be compatible? If there is already a solution in this thread, would you please link it? :-)

proposals/UIWebView-strategy.md
breautek: Just to play devil's advocate here, on the topic of local storage...

There are some plugins out there that migrate local storage such as [this](https://www.npmjs.com/package/cordova-plugin-migrate-localstorage) one, but most of them will only migrate if the `WkWebView` storage containers are empty (otherwise they run into a question of "should they overwrite?").

If one would update cordova-ios and thus migrate to the wkwebview without the use of this plugin, they may not realise that they lost data, and by that time it's too late to use a migration plugin. To save headaches for developers, maybe we should incorporate of these plugins directly into cordova-ios.

But on the other side, we could just simply tell users to consider using one of those plugins so that we don't have to maintain an one-time use code for an entire major version lifetime.

Personally I think I'm still on the "not cordova's problem" side...

breautek:
>  also haven't seen the solution yet in the thread.

The current solution, which is available in `cordova-ios@5.1.1` & if you're using the inappbrowser, `cordova-plugin-inappbrowser@3.2.0` makes use of a flag to hide the UIWebView references behind a compile-time flag. This with a WkWebView plugin, either cordova official plugin, or ionic's webview plugin (ionic webview can be used without using the ionic framework), is a current day solution to Apple's deprecation notice.

`cordova-ios@6` (which is a work in progress) will completely remove the `UIWebView` code and essentially make the `WkWebView` the default webview, eliminating the need to use a wkwebview plugin.

> or do we have to use the WKWebView plugin and re-write our code completely to be compatible

Regardless due to the reasons above, you're going to have to move away from the UIWebView, and migrate to WkWebView eventually. No way around that. Even if you're writing a pure native ios app, depending on the features you use, this might be breaking changes for your app. To name a couple, if you use web storage, migrating webviews does not bring over web storage to the new webview. WkWebView enforces [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS), so that may require backend changes if you make any requests to a server to be CORS compatible. For the web storage issue, there are some plugins to help migrate web storage such as [this](https://www.npmjs.com/package/cordova-plugin-migrate-localstorage) one (not an endorsement, never actually used this plugin).

> If there is already a solution in this thread, would you please link it? :-)

https://github.com/apache/cordova-ios/pull/715

ragcsalo:
Many-many thanks!!!!! :-)

HansKrywalsky:
> The current solution, which is available in `cordova-ios@5.1.1` & if you're using the inappbrowser, `cordova-plugin-inappbrowser@3.2.0` makes use of a flag to hide the UIWebView references behind a compile-time flag. This with a WkWebView plugin, either cordova official plugin, or ionic's webview plugin (ionic webview can be used without using the ionic framework), is a current day solution to Apple's deprecation notice.
> 
> `cordova-ios@6` (which is a work in progress) will completely remove the `UIWebView` code and essentially make the `WkWebView` the default webview, eliminating the need to use a wkwebview plugin.
> 
> [apache/cordova-ios#715](https://github.com/apache/cordova-ios/pull/715)

I'm going to develop a complete new App for IOS. Will i have any Problems migrating from cordova-ios@5.1.1 to 6.0.0? Regarding Storage lost, etc.?

breautek:
> I'm going to develop a complete new App for IOS. Will i have any Problems migrating from cordova-ios@5.1.1 to 6.0.0? Regarding Storage lost, etc.?

If you're developing a brand new app, you wouldn't have any data stored for your users in the first place. But if you make ajax requests to remote server, CORS could still be an issue, that is an entirely a backend thing though. If you're unfamiliar with cors, I suggest you read more on it at https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS

breautek:
Sorry I misread your question. web storage issues happens when you change webviews, so going from `UIWebView` to the `WKWebview`.

So if you're already on `WKWebView` using a plugin, on cordova-ios@5.1.1, upgrading to 6.0.0 shouldn't cause any storage migration issues.

If you're not using any wkwebview plugins, and you upgrade from 5.1.1 to 6.0, then you definitely will.

jcesarmobile:
> So if you're already on `WKWebView` using a plugin, on cordova-ios@5.1.1, upgrading to 6.0.0 shouldn't cause any storage migration issues.

well, it really depends, if we implement the `WKUrlSchemeHandlers` as proposed here, the app won't be served from file, but from a new custom scheme and data will be lost.

HansKrywalsky:
@jcesarmobile So what so you prefer? Wait for 6.0.0? When is it going to be released?

ragcsalo:
There are already nightly builds of 6.0.0 :-) 
Also release 5.1.1 provides a preprocessor directive that allows apps to “compile out” UIWebView code.

Are there any tutorials on migrating to WKWebView?

thealiano:
Any updates for the 6.0.0 ?

ragcsalo:
2 weeks left... I hope there will be an easy and quick solution to migrate existing apps to WKWebView 🤔

NiklasMerz:
We just published a blog post with steps to take now and a note about the future roadmap.

https://cordova.apache.org/howto/2020/03/18/wkwebviewonly.html

I am locking this now since the plan is clear and to avoid more questions on this thread. Other committers can still comment and open this again if needed.

erisu:
As we have completed the removal of UIWebView from the iOS platform, I think we can either close or merge this PR.

I believe most proposals are usually merged before the task is started or completed.

